PR #3, opened 2015-10-20
Added grunt to dev dependencies

Soviut:
Relates to https://github.com/dmlap/generator-videojs-plugin/issues/2


misteroneill:
Thanks! We are currently undergoing a rewrite of this project to be bigger and better, but this will make a useful patch for 0.1.x, I think.

PR #5, opened 2015-11-02
Rewrite

misteroneill:
This project has been languishing for a while, but with our new effort to standardize plugin authoring practices, it's a good time to bring it back to life.

This PR represents a complete rewrite of the generator to generate video.js 5.0-compatible plugins, which use ES6, Browserify, and more modern goodies by default.


misteroneill:
Just tossing out a random thought here, but the generator is currently being rewritten to always support npm-based workflows, but with optional Grunt tasks behind them. I've done a few tests and it seems - to my surprise - that the Grunt-based builds are _considerably faster_ than the npm-based builds. The extra time seems to be on the npm side.

- `grunt build` is ~1.8s
- `npm run build` (alias for `grunt build`) is ~2.5s
- `npm run build` is ~6.3s

I still think the idea of having npm scripts available to run these commands is critical (for abstracting away the underlying build system, for the use of CI servers, etc), but I'm wondering if we ought to expend the energy maintaining both a Grunt-based and an npm-based set of scripts in the generator.

misteroneill:
I added [an experimental branch](https://github.com/videojs/generator-videojs-plugin/compare/rewrite...grunt-only) that removes the npm build scripts and leaves only the Grunt setup.

misteroneill:
Upon further investigation, the build is slower, but watch tasks are faster when using the npm-based tooling - and these tasks seem more performance-sensitive. Going to stick with the original plan and ditch the aforementioned experimental branch.

misteroneill:
I think this is ready for review. One thing I'm thinking is moving the contents of `test/unit/` to `test/`.

.travis.yml
```diff
@@ -1,7 +1,6 @@
+sudo: false
 language: node_js
 node_js:
+  - 'iojs'
```
gkatsev: should probably be `stable` or something or `4.x` instead.

.jshintrc
```diff
@@ -11,11 +11,7 @@
...
   "noarg": true,
   "quotmark": "single",
-  "regexp": true,
   "undef": true,
   "unused": true,
-  "strict": true,
-  "trailing": true,
-  "smarttabs": true,
-  "white": true
+  "strict": true
```
gkatsev: shouldn't we use `videojs-standard` to lint this project as well? :D

generators/app/package-json.js
```diff
@@ -0,0 +1,268 @@
...
+        'karma-detect-browsers': '^2.0.0',
+        'karma-firefox-launcher': '^0.1.0',
+        'karma-ie-launcher': '^0.2.0',
+        'karma-opera-launcher': '^0.3.0',
+        'karma-qunit': '^0.1.0',
+        'karma-safari-launcher': '^0.1.0',
+        'npm-run-all': '^1.2.0',
+        'qunitjs': '^1.0.0',
+        'sinon': '^1.0.0',
+        'video.js': '^5.0.0',
```
gkatsev: if we're doing the split build as in https://github.com/brightcove/videojs-dock/pull/3
`video.js` would need to be in `dependencies` instead.

generators/app/package-json.js
```diff
@@ -0,0 +1,268 @@
...
+        'build:test',
+        'clean',
+        'start',
+        'test',
+        'watch',
+        'watch:js',
+        'watch:test'
+      ]),
+      devDependencies: {
+        'grunt': '^0.4.0',
```
gkatsev: we should add [`time-grunt`](https://www.npmjs.com/package/time-grunt) because it's useful.

generators/app/package-json.js
```diff
@@ -0,0 +1,268 @@
...
+        'build:js': 'npm-run-all mkdirs browserify:js bannerize:js uglify',
+        'build:test': 'npm-run-all mkdirs browserify:test',
+        'clean': 'rm -rf dist test/unit/dist && npm run mkdirs',
+        'mkdirs': 'mkdir -p dist test/unit/dist',
+        'prestart': 'npm-run-all -p docs build',
+        'start': 'npm-run-all -p serve watch',
+        'serve': 'babel-node scripts/server.js',
+        'pretest': 'npm-run-all lint build:test',
+        'test': 'karma start test/karma/detected.js',
+        'watch': 'npm-run-all mkdirs -p watch:js watch:test',
```
gkatsev: `npm-run-all -p watch:*`

generators/app/templates/_.npmignore
```diff
@@ -0,0 +1 @@
+!dist/
```
gkatsev: would this make npm only package up what is in `dist/`? If so, we probably should just include everything in the package instead.

generators/app/templates/docs/_standards.md
```diff
@@ -0,0 +1,126 @@
...
+
+npm Script   | Grunt Equiv.       | Optional | Description
+------------ | ------------------ | -------- | -----------
+`build`      | `grunt build`      |          | Runs all build sub-tasks.
+`build:css`  | `grunt build:css`  | ✓        | Builds the Sass entry point.
+`build:js`   | `grunt build:js`   |          | Builds the Browserify entry point.
+`build:test` | `grunt build:test` |          | Builds the test Browserify entry point.
+`clean`      | `grunt clean`      |          | Cleans up build artifacts.
+`lint`       | `grunt lint`       |          | Lints all `.js` file(s).
+`start`      | `grunt start`      |          | Starts a development server and runs `watch`.
```
gkatsev: We should perhaps point out that `start` and `test` are default scripts in npm. So, running those only requires `npm test` or `npm start` while the others are `npm run build`.
Also, `npm run` will list all of the scripts that are available.

generators/app/templates/scripts/_server.js
```diff
@@ -0,0 +1,23 @@
...
+const app = connect();
+
+app.use(serveStatic(path.join(__dirname, '..')));
+
+portscanner.findAPortNotInUse(PORT, MAX_PORT, HOST, (error, port) => {
+  if (error) {
+    throw error;
+  }
+
+  app.listen(port);
```
gkatsev: we should make sure to log what port we're using.

package.json
```diff
@@ -1,36 +1,35 @@
 {
   "name": "generator-videojs-plugin",
-  "version": "0.1.2",
-  "description": "Yeoman generator for video.js plugins",
-  "license": "MIT",
-  "main": "app/index.js",
-  "repository": "dmlap/generator-videojs-plugin",
+  "version": "0.0.0",
+  "description": "A Yeoman generator for VideoJS plugins",
+  "license": "Apache 2.0",
```
gkatsev: Should be `Apache-2.0`

package.json
```diff
@@ -1,36 +1,35 @@
...
-  },
-  "engines": {
-    "node": ">=0.10.0"
+    "name": "VideoJS",
+    "url": "https://github.com/videojs"
   },
   "scripts": {
     "test": "mocha"
   },
   "files": [
```
gkatsev: can we package up everything instead of just `generators/`?

package.json
```diff
@@ -1,36 +1,35 @@
...
   },
   "devDependencies": {
-    "mocha": "*"
-  },
-  "peerDependencies": {
-    "yo": ">=1.0.0"
+    "karma-chrome-launcher": "^0.2.1",
+    "karma-firefox-launcher": "^0.1.6",
+    "karma-ie-launcher": "^0.2.0",
+    "karma-qunit": "^0.1.8",
```
gkatsev: do we even need karma in this project?

gkatsev: Is this to test that our generator is working and that the generated tests are working?

gkatsev:
:+1: on moving `tests/unit` to `test/`. Have a few small comments that should be fixed before merging.
The work to split the `dist` and `src` as in brightcove/videojs-dock#3 could be done after merging as another PR.

.jshintrc
misteroneill: Yeah, well, I didn't care to look into how we might write a Yeoman generator in ES6. In theory, I agree. :)

package.json
misteroneill: That's... a mistake. I have no idea how it happened! #CodingWhileSick

.jshintrc
gkatsev: Move all the generator code to `src/` and then compile using babel into `generator/` on `npm publish`. Or something. I'm ok with holding off on doing this.

.travis.yml
misteroneill: Good catch. Updating to `stable`, `4.2`, `0.12`, and `0.10`.

gkatsev: Probably no need for `0.12`

package.json
misteroneill: Sure, this came from the generator-generator, I think. I'll remove it.

package.json
misteroneill: :+1:

generators/app/templates/scripts/_server.js
misteroneill: :+1:

generators/app/templates/docs/_standards.md
misteroneill: I'm cool with noting the first part, but don't feel that we should document that `npm run` will list available scripts.

generators/app/templates/_.npmignore
misteroneill: This is to override `.gitignore`'s exclusion of `dist/` - I think it means everything would be included.

generators/app/package-json.js
misteroneill: :+1:

generators/app/templates/_.npmignore
gkatsev: If I understand npmignore correctly, it means that only `dist/` will be included in the package for the module. I would like to include everything -- docs, sources, libs, tests -- and not just `dist/`.

generators/app/templates/docs/_standards.md
gkatsev: We should just link to npm's documentation.

gkatsev:
Another feature request: use [doctoc](https://www.npmjs.com/package/doctoc) to generate ToCs in the READMEs.

generators/app/templates/_.npmignore
misteroneill: I don't think it means that. Otherwise, the advice for overriding `.gitignore` wouldn't be to create an empty `.npmignore`.

Also, `npm install -g .` on a generated plugin includes everything (with `dist/`).

misteroneill:
I'm not entirely convinced of the usefulness of doctoc, but I'm willing to consider it. There does appear to be a Grunt plugin; though, perhaps it's something that's best left as an npm script (perhaps part of the `preversion` script).

gkatsev:
It works great for automatically generating a TOC automatically. https://github.com/brightcove/videojs-playlist#api as an example, though, I did that one by hand before finding doctoc.
Regardless, can be a later addition.

misteroneill:
I get what it does; I just find it a bit clunky, I guess. But there's only so much you can do with a static markup file. I suppose I'm alright with including it as an optional feature (i.e. there's a Yeoman prompt for it). Added to the TODOs above.

misteroneill:
After talking through some plugin dependency issues at work, I realize we really need more conventions around where plugins cache their data and what method(s) they should expose as far as lifecycle (setup, teardown, etc).

gkatsev:
Yep, that's something I want to work on for a videojs-next. Define more structure to plugins. Maybe like how popcornjs works or who knows. Rigtht now plugins are both super simplistic but also super powerful because they just get the player.

gkatsev:
I wonder, can we have this module depend on `yo` and then export a bin called `videojs` or something that basically is basically a shortcut for `yo generator-videojs-plugin`, so, you could just do
```js
$ npm i -g videojs-plugin
$ videojs
```

misteroneill:
Ah, interesting idea! I like it (no implied dependency on global `yo`).

misteroneill:
There doesn't seem to be a nice way to get this working without requiring global `yo`, but I've got something that's most of the way there. The steps will be:

```sh
$ npm i -g generator-videojs-plugin
$ vjsp
```

`yo@^1.5.0` is now a peer dependency of the generator; so, it should be installed alongside the generator. I don't think that'll be too bad...

gkatsev:
In npm 3, peerDeps are no longer installed by default.

misteroneill:
Well, crap. I guess I didn't read carefully enough. Then it's...

```sh
$ npm i -g yo generator-videojs-plugin
$ vjsp
```

forbesjo:
Should the scripts be written using shelljs so they can work on Windows?

misteroneill:
Certainly not for the initial version. The Grunt tasks might just work out of the box (perhaps with some modification). I'm not sure what dependencies may or may not work on Windows.

misteroneill:
:wave: Bye, Grunt!

LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright Brightcove, Inc.
```
gkatsev: We probably should apply [the new videojs governance and licensing](https://github.com/videojs/video.js/wiki/Governance#licensing-and-ownership) to all repos in the org. We should also make it more formal and actually put it into the videojs repo and what not.
Not sure exactly how this particular project should be copyrighted.

.travis.yml
```diff
@@ -1,7 +1,7 @@
+sudo: false
 language: node_js
 node_js:
+  - 'stable'
```
gkatsev: https://github.com/creationix/nvm/issues/870 something to think about. We might want to change `stable` to `node`. And at some point we'll be able to change `4.2` to `lts` or `argon` or something.

docs/standards.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Packaging and Dependencies
+
+__All standard video.js plugins must be npm packages.__
+
+Plugins should be as self-contained as possible, so the only dependency (`"dependencies"` in `package.json`) by default is video.js, which is shimmed using `browserify-shim`.
+
+Development dependencies (`"devDependencies"` in `package.json`) will include many packages related to developing, building, and testing a standard video.js plugin.
+
+### Structure
```
gkatsev: Should we add a column for "generated"? All folders and files that are generators automatically during builds.

src/app/package-json.js
```diff
@@ -0,0 +1,303 @@
...
+ *    "postversion": "..."
+ *
+ * @param  {Object} source
+ * @return {Object}
+ *         A new ordered object.
+ */
+const alphabetizeScripts = (source) => {
+  let keys = Object.keys(source);
+  let prePost = keys.filter(
+    k => _.startsWith(k, 'pre') || _.startsWith(k, 'post')
```
gkatsev: If you include `es6-shim`, you could just do `k.stratsWith('pre')` :P

src/app/templates/_.gitignore
```diff
@@ -0,0 +1,38 @@
...
+npm-debug.log*
+
+# Dependency directories
+bower_components/
+node_modules/
+
+# Yeoman meta-data
+.yo-rc.json
+
+# Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
```
gkatsev: Do we need to ignore grunt since we're no longer using it?

src/app/templates/_.npmignore
```diff
@@ -0,0 +1,4 @@
+!docs/api/
```
gkatsev: Why just `docs/api`?

src/app/templates/_.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: node_js
+node_js:
+  - 5
```
gkatsev: as mentioned above, we probably want `node` and `4.2` to get the latest supported stable version and latest LTS version.

src/app/templates/_bower.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "<%= nameOf.package %>",
+  "version": "0.0.0",
```
gkatsev: We should follow `npm`s default of 1.0.0, I think. (Though, it does make it weird trying to version and publish your first release).

src/app/templates/test/_plugin.test.js
```diff
@@ -0,0 +1,54 @@
...
+QUnit.test('the environment is sane', function(assert) {
+  assert.strictEqual(typeof Array.isArray, 'function', 'es5 exists');
+  assert.strictEqual(typeof sinon, 'object', 'sinon exists');
+  assert.strictEqual(typeof videojs, 'function', 'videojs exists');
+});
+
+QUnit.module('<%= nameOf.package %>', {
+
+  beforeEach() {
+    this.player = new Player(document.createElement('video'));
```
gkatsev: This element should probably get attached to the qunit fixture element.

Also, we probably should move the test helpers from videojs into a separate module or make it be requireable.

src/app/templates/_.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: node_js
```
gkatsev: we should add `sudo: false` here as well.

src/app/templates/_bower.json
misteroneill: Yeah, that's why I used 0.0.0 - because then you can't (easily) use `npm version major` for your first release.

.travis.yml
misteroneill: :+1:

docs/standards.md
misteroneill: Done! Also added a bit of a legend after the table.

src/app/package-json.js
misteroneill: So much lodash hate! :stuck_out_tongue: Yeah, I think that's a better approach.

src/app/templates/_.gitignore
misteroneill: Nope.

src/app/templates/_.npmignore
misteroneill: Good question.

src/app/templates/test/_plugin.test.js
misteroneill: That makes sense. I'll move the element.

We should definitely address the test helpers, but not in this PR. :)

src/app/package-json.js
misteroneill: But lodash is used all over the place, so not much point. Maybe this can be a future refactor.

src/app/templates/_.travis.yml
misteroneill: Done.

src/app/package-json.js
gkatsev: Yeah, totally not required or any action is necessary :)

src/app/templates/_bower.json
gkatsev: The issue is that then you end up with a lot of plugins are basically permanently in the 0.x zone. Which is why npm chose to change the default.

README.md
```diff
@@ -1,50 +1,98 @@
...
+- [License](#license)
 
-![](http://i.imgur.com/JHaAlBJ.png)
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-Basically, he wears a top hat, lives in your computer, and waits for you to tell him what kind of application you wish to create.
+## Getting Started
 
-Not every new computer comes with a Yeoman pre-installed. He lives in the [npm](https://npmjs.org) package repository. You only have to ask for him once, then he packs up and moves into your hard drive. *Make sure you clean up, he likes new and shiny things.*
+If you don't know what Yeoman is or what generators are, check out the Yeoman [Getting Started][getting-started] document. Long story short, make sure you have Yoeman and this generator installed globally:
```
hartman: Yoeman

src/app/templates/_README.md
```diff
@@ -0,0 +1,40 @@
+# <%= nameOf.package %>
+
+A plugin for video.js.
```
hartman: Lost the insertion of the plugin description here, compared to the previous version.

misteroneill: Thanks!

src/app/templates/_.travis.yml
forbesjo: Aren't all new projects on Travis on the container-based infrastructure by default?

forbesjo: Actually this would be useful when remaking a project.

gkatsev: Not positive but it's probably safer to keep it.

README.md
misteroneill: Thanks, good catch!

misteroneill:
Once we have a new version of `videojs-standard` published to npm, this can be merged and published as 1.0.0. However, we are blocked by an open question about the existence of unused tags left over from when `videojs-standard` was a fork.

PR #16, opened 2015-12-09
Fix npm packaging

misteroneill:
This makes it based on the generated `.gitignore` with the difference that build directories are _not_ ignored by NPM as they are needed for projects that depend on them.

### DIFF
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -107,8 +107,13 @@ const packageJSON = (current, context) => {
       ]
     },
     'files': [
-      'dist/**/*.*',
-      'es5/**/*.js'
+      'dist/',
+      'dist-test/',
+      'es5/',
+      'scripts/',
+      'src/',
+      'test/',
+      'index.html'
     ],
     'scripts': {
       'prebuild': 'npm run clean',
@@ -214,6 +219,8 @@ const packageJSON = (current, context) => {
 
   if (context.isPrivate) {
     result.private = true;
+  } else {
+    result.files.push('CONTRIBUTING.md');
   }
 
   // Create scripts for each Karma browser.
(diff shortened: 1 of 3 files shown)


src/app/package-json.js
```diff
@@ -268,6 +272,8 @@ const packageJSON = (current, context) => {
 
   // Support the documentation tooling option.
   if (context.docs) {
+    result.files.push('docs');
```
misteroneill: This may or may not make sense. Probably not, honestly.

misteroneill:
Blocked by virtue of discussion in #18

src/app/package-json.js
gkatsev: Yes, docs should be in there. Everything should.

src/app/package-json.js
```diff
@@ -106,6 +106,10 @@ const packageJSON = (current, context) => {
         'scripts'
       ]
     },
+    'files': [
+      'dist/**/*.*',
+      'es5/**/*.js'
```
gkatsev: The original source files should also be included.

src/app/package-json.js
misteroneill: Why? It seems to me the package should only require those things needed to use it.

src/app/package-json.js
gkatsev: Also, test files.

src/app/package-json.js
gkatsev: Because it makes the npm tarball self contained and makes the docs for that version available locally and offline.
Also, there's really no reason to *not* add it to npm.

misteroneill: Alright, I can see the case for `docs/`. But why would the `src/` or `tests/` need to be there? Seems to me it just adds pointless files to the package. Everything is transpiled by babel into `es5/`, so those two directories serve no purpose in the package.

gkatsev: Having the `src` in there allows users who want to do their own compilation the option to do so (most commonly webpack users). Having both `src` and `tests` in the tarball also makes it possible to completely recover a repo from the npm package if necessary.

misteroneill: Fair enough. I'm good with that. :+1:

gkatsev:
LGTM

PR #17, opened 2015-12-09
Use node-static

misteroneill:
Rather than the manually-written `server.js` to serve static files, this simplifies things to use the `node-static` package. The only thing this changes is the auto-port-detection behavior (i.e. if 9999 is not available, use 10000... and so on).


misteroneill:
Marking this as 2.0.0 because I guess it's technically a breaking change.

misteroneill:
I'm going to kill this one because I think it'd be better to do something using [`budo`](https://www.npmjs.com/package/budo).

PR #31, opened 2015-12-30
added browserify debug to generate source maps during karma test runs

BrandonOCasey:
I found myself adding this to nearly all of the plugins that I have been working on, because it is not very helpful for errors to point to a browserify conversion, when they should really point me to the location in the source where something is wrong.


misteroneill:
Cool. I'll put this in the next milestone. :+1:

gkatsev:
fwiw, I tend to disable these because they generally cause more issues than they solve. Unfortunately.
Probably still worth having it in, though.

misteroneill:
This piece is actually going to be part of `spellbook`; so, I'm going to close this. But thanks for the submission. I'll see if we can get useful source maps in there. :+1:

PR #32, opened 2015-12-30
Fixes for package.json

### LINKED ISSUE #24: When bower is enabled the dist directory is not included in npm packages
Normally, when you do `npm version && npm publish`, your dist ~~module~~ directory is included in the package. When you generated a package that has bower enabled, it seems that I'm not able to do that.

I need to do `npm version && npm run build && npm publish`.

### LINKED ISSUE #26: Ordering of name, description, version, main is moved when using the generator
The order of those 4 fields should stay the same/at the top of the package.json file.

### LINKED ISSUE #29: Better handling of keywords field in package.json.
Right now, this gets blown away if applied to any projects. There should be some merging here.

misteroneill:
This does several things:

- [x] Fixes #24. Always build before publishing. This is necessary because the Bower-friendly versioning process will blow away the dist/ directory.
- [x] Fixes #26. Ordering of existing `package.json` is preserved and certain properties are not carried over.
- [x] Fixes #29. Keywords are now merged instead of being blown away.


gkatsev:
Do we have tests for the generator? That would be my only comment.

misteroneill:
Yeah, but not great tests for modifying existing packages. I'll have to do some reading on how to achieve that with Yeoman.

gkatsev:
Alright, not sure it should up this PR, though.

misteroneill:
Yeah, that was my thought.

misteroneill:
On the other hand, I think I want to get at least a basic test in there before merging this change. It seems like it's not too complicated.

misteroneill:
I added a few tests related to this. At first, I tried running the generator multiple times against a single file system, but that was proving to be overly complex. So, I think directly testing the merging behaviors of the `package-json` module should suffice.

gkatsev:
LGTM

Issue #30, opened 2015-12-30
Don't push during the version process.


misteroneill:
There is such a thing as being too clever and magical and this is one of those.

Right now, both `version` processes will automatically push commit(s) the created tag for you. This should be a separate/explicit step, like publishing to npm.

1. Create commit(s) and tag(s) for next version (`npm version *`).
1. Push commit(s) and tag(s) to repository (`git push && git push --tags`).
1. Publish to npm (`npm publish`).

Issue #28, opened 2015-12-30
Stop using documentation project.


misteroneill:
The [documentation](http://documentation.js.org/) project offers some nice things (e.g., Markdown), but it does something really unbelievable: it flattens the structural representation of classes and modules; so, every function/symbol is given the same hierarchical weight.

We should use [jsdoc3/jsdoc](https://github.com/jsdoc3/jsdoc) by itself instead.

misteroneill:
Maybe rather than switching to JSDoc, we ought to not suggest any documentation standard for video.js plugins?

PR #33, opened 2016-01-02
Windows

### LINKED ISSUE #7: Windows
Support Windows. Work is ongoing in #33

In the generator itself:

- [x] `"prebuild"`
- [x] `"build:generator:copy"`
- [x] ~~`vjsplugin`~~ Removed.

In generated projects:

- [x] `"mkdirs"`
- [x] `"clean"`
- [x] `"build:test:browserify"`
- [x] `"build:css:sass"`
- [x] `"docs"`
- [x] `"watch:css"`
- [x] `"watch:test"`
- [x] ~~`scripts/npm-preversion-for-bower.sh`~~ Removed.
- [x] `scripts/npm-version-for-bower.js` Renamed.
- [x] `scripts/npm-postversion-for-bower.js` Renamed.

misteroneill:
This makes the generator and generated projects work on Windows via shelljs. Unfortunately, it doesn't currently expose nice executables; so, I am making unsightly use of `node -e` in a few places.

Fixes #7, #28, and #30


package.json
```diff
@@ -23,13 +22,12 @@
     ]
   },
   "scripts": {
-    "prebuild": "npm run clean",
+    "prebuild": "node -e \"var s=require('shelljs'),d=['generators','es5-check','es5-test'];s.rm('-rf',d);s.mkdir('-p',d)\"",
```
forbesjo: I made it a little cleaner by making a clean.js file:
```js
var s = require('shelljs'),
  d = process.argv.slice(2);
s.rm('-rf', d);
s.mkdir('-p', d);
```
` "prebuild": "shjs clean.js generators es5-check es5-test",`

misteroneill: Yeah, I was honestly hoping to avoid creating new `.js` files because they'd be repeated on future projects making it harder to remove automatically in the future in projects based on the generator. The `package.json` scripts are easier to update.

gkatsev: The generator could detect if those files exist and ask the user whether they'd like the files removed because they are no longer necessary. Or something.

misteroneill: I suppose that's true. There could be a generic prompt like "Remove files that are no longer needed? (Y/n)"

gkatsev: does `node -e` accept single quotes? Might be able to not need to escape them then.

misteroneill: But then wouldn't we need to escape the quotes inside the script?

gkatsev: Oh, right...
In that case, run it through `babel-node` and then just use a backtick for strings? :P

misteroneill: The other thing here is that I fully intend a future release to change these `node -e` scripts to something much cleaner, but I haven't had time to write a PR for shelljs or to write a CLI wrapper package.

misteroneill:
Realized that the updated versioning process here for Bower missed a big piece: bumping the version number in `bower.json`. Will have to fix that before merging.

misteroneill:
_Nevermind_. `bower.json` is no longer supporting the `"version"` field. Good riddance!

package.json
gkatsev: This won't be an issue once it's part of the [spellbook](https://github.com/videojs/spellbook) ( #42 )

misteroneill: Yep!

src/app/templates/scripts/_watch-test.js
```diff
@@ -0,0 +1,23 @@
+var browserify = require('browserify');
+var fs = require('fs');
+var glob = require('glob');
+var watchify = require('watchify');
+
+glob('test/**/*.test.js', function(err, files) {
+  var b = browserify(files, {
```
gkatsev: would be great when we update to spellbook/budo.

gkatsev:
LGTM

src/app/templates/scripts/_watch-test.js
misteroneill: Yeah, it's gonna save so much cruft. :+1:

PR #34, opened 2016-01-05
added babelify transform to browserify transforms in package.json

BrandonOCasey:
In order for browserify to know that the module we are trying to use requires babelify to be run (so that we can transform the es6 code to es5, and then do normal browserify things) we have to include babelify as a transform in the package.json. If we do not include this then things will not be able to require our modules with browserify because they will not be babelified at all, and browserify will complain about syntax issues when you go to try.

### DIFF
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -165,7 +165,7 @@ const packageJSON = (current, context) => {
     'license': context.nameOf.license,
 
     'browserify': {
-      transform: ['browserify-shim']
+      transform: ['browserify-shim', 'babelify']
     },
 
     'browserify-shim': {


misteroneill:
Actually, this was left out intentionally. The reason is that there are two compile steps from the source JS.

First, Babel transpiles `src/` into `es5/` for npm and Browserify consumers. This becomes the source of truth for the module (i.e. if you `npm install videojs-foo`, then `import foo from 'videojs-foo'`, it should use `es5/plugin.js`). This happens via the `"main": "es5/plugin.js"` in `package.json`.

Next, Browserify builds `es5/` into `dist/` (which has already been transpiled by Babel, so no need to use the babelify transform). This is the stand-alone distributable for consumers using `<script>` tags (or their favorite front-end script loader).

PR #39, opened 2016-01-11
Address author prompt from object

### LINKED ISSUE #10: Author prompt should deal with objects.
As an example, if `author` is set as an object, the prompt still appears with `[object Object]`. Nothing _breaks_ but it'd be nice to think through this experience a little better.

misteroneill:
This resolves the situation where an existing package.json has an author field, but it is an object instead of a string. In this case, the prompt is skipped.

Also includes some code cleanup that doesn't change behavior in any way.

Fixes #10


incompl:
LGTM

PR #43, opened 2016-01-14
Add scoped package support

### LINKED ISSUE #40: Support scoped packages
At Brightcove, we use scoped npm packages. We should have a prompt (and a `--bcov` default) for this possibility.

misteroneill:
Pretty self-explanatory. Also, some related code cleanup and efforts toward DRYness. 

Fixes #40

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -46,13 +46,17 @@ export default yeoman.generators.Base.extend({
   },
 
   /**
-   * Gets the name of the scope including the "@" symbol.
+   * Gets the name of the scope including the "@" symbol. Will not prepend
+   * the "@" if it is already included.
    *
    * @param  {String} scope
    * @return {String}
    */
   _getScope(scope) {
-    return scope && typeof scope === 'string' ? `@${scope}` : '';
+    if (!scope || typeof scope !== 'string') {
+      return '';
+    }
+    return scope.charAt(0) === '@' ? scope : `@${scope}`;
   },
 
   /**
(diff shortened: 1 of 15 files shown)


src/app/index.js
```diff
@@ -59,6 +46,70 @@ export default yeoman.generators.Base.extend({
   },
 
   /**
+   * Gets the name of the scope including the "@" symbol.
+   *
+   * @param  {String} scope
+   * @return {String}
+   */
+  _getScope(scope) {
+    return scope && typeof scope === 'string' ? `@${scope}` : '';
```
gkatsev: should this make sure you don't get double `@` or is there extra checking for that elsewhere?

gkatsev:
Will it prefix `videojs-` to scoped packages as well?

gkatsev:
Related to my last question, do we have an escape hatch mechanism if you really wan to use generator but don't want it prefixed with `videojs-`?

misteroneill:
1. Yes, providing a scope results in `@SCOPE/videojs-PLUGIN`
1. Not currently. I feel like it's a convention we ought to strongly encourage. That said, there's nothing stopping anyone from changing after the generator runs.

src/app/index.js
misteroneill: It's enforced in the [validation for the prompt input](https://github.com/videojs/generator-videojs-plugin/pull/43/files#diff-22e911e52dc237dc1da62b1051879718R39), but I think it would make sense to double-check (since prompts can be skipped, etc).

misteroneill:
@gkatsev Addressed the potential for `_getScope` to double the `@` in a scope even though it was protected during the prompting process. Any more comments?

gkatsev:
LGTM

PR #46, opened 2016-01-26
use global 'browser' file set in package.json

BrandonOCasey:
set standalone build name to something that makes sense

### DIFF
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -86,6 +86,7 @@ const packageJSON = (current, context) => {
     'version': context.version,
     'description': context.description,
     'main': 'es5/plugin.js',
+    'browser': 'src/plugin.js',
 
     'scripts': _.assign({}, current.scripts, {
       'prebuild': 'npm run clean',
@@ -107,7 +108,10 @@ const packageJSON = (current, context) => {
       ]),
 
       'build:js:browserify': scriptify([
-        'browserify . -s %s -o dist/%s.js'
+        'browserify .',
+        '-t babelify',
+        '-s ' + context.functionName,
+        '-o dist/%s.js'
       ]),
 
       'build:js:uglify': scriptify([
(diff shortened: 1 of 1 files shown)


src/app/package-json.js
```diff
@@ -86,6 +86,7 @@ const packageJSON = (current, context) => {
     'version': context.version,
     'description': context.description,
     'main': 'es5/plugin.js',
+    'browser': 'src/plugin.js',
```
gkatsev: since we don't configure browserify to run the `babelify` transform, we actually don't want to do this. Browserify would then try to pick up the src and that's es6 code and it'll fail.

src/app/package-json.js
```diff
@@ -107,7 +108,10 @@ const packageJSON = (current, context) => {
       ]),
 
       'build:js:browserify': scriptify([
-        'browserify . -s %s -o dist/%s.js'
+        'browserify .',
+        '-t babelify',
```
gkatsev: browserify should depend on running the babel compiler instead of specifying the transform here.

src/app/package-json.js
```diff
@@ -107,7 +108,10 @@ const packageJSON = (current, context) => {
...
+        '-s ' + context.functionName,
```
gkatsev: Is `context.pluginName` not what we want the standalong name to be? How does `context.functionName` compare?

misteroneill: I was about to ask this very question. I'm not sure how `functionName` "makes sense" and `pluginName` doesn't...

src/app/package-json.js
misteroneill: Yes, it's a two-step process because we need both the transpiled code (in `es5/`) and the Browserified output (in `dist/`).

src/app/package-json.js
BrandonOCasey: it seems I had a previous version of the generator installed that was setting standalone to something like this: 
browserify -s src/stub.js

BrandonOCasey:
appear that I just had an outdated generator

PR #48, opened 2016-01-30
Simplify build:css, fix watch:css.

misteroneill:
This is a backward-compatible simplification of the `build:css` task and a bug fix for the incorrect `watch:css` task.

PR #50, opened 2016-01-30
Karma improvements, remove shelljs, vjsplugincheck improvements

misteroneill:
This removes shelljs from generated packages because it is no longer needed after #48 was merged as 1.0.4. Also, this brings those changes into the pre-2.0.0 history. Instead, `rimraf` and `mkdirp` are used where those features were still needed (i.e. the `"clean"` script).

The second change here is to dramatically simplify the Karma setup and move compiled tests from `dist-test/` to `test/dist/`. I was trying to tab-complete `dist/` too often and it was getting annoying. :)

More importantly, there is now only one Karma config file generated. This takes advantage of the fact that the command line arguments _are provided_ in the `configs` object passed to the `karma.conf.js` module! I have tested and verified that the new `karma.conf.js` works in a generated project.


gkatsev:
LGTM

PR #51, opened 2016-02-02
Add CHANGELOG support

misteroneill:
This is a pretty barebones implementation, but it automates some of the tedium out of maintaining a CHANGELOG. It also makes the CHANGELOG optional (on by default).


docs/standards.md
```diff
@@ -120,24 +120,26 @@ __All standard video.js plugins must implement the core set of npm scripts.__
...
-`watch:test` |          | Triggers a build when the test entry point changes.
-`version`    | ✓        | Special handling if Bower support is enabled.
+npm Script    | Optional | Description
+------------- | -------- | -----------
+`build`       |          | Runs all build sub-tasks.
+`build:css`   | ✓        | Builds the Sass entry point.
+`build:js`    |          | Builds the Browserify entry point.
+`build:lang`  | ✓        | Builds language files.
+`build:test`  |          | Builds the test Browserify entry point.
+`change`      | ✓        | Adds a free-form text entry to the CHANGELOG.
```
gkatsev: would you run it as:
```sh
$ npm run change -- this is the item to add to the changelog
```
?

src/app/templates/scripts/_version.js
```diff
@@ -0,0 +1,68 @@
...
+    return false;
+  }
+};
+
+var commands = [];
+
+// If the project has a CHANGELOG, update it for the new release.
+if (hasChangelog()) {
+  commands = commands.concat([
+    'chg release "' + pkg.version + '"',
```
gkatsev: this would get the new version we want?

docs/standards.md
misteroneill: You can do that, but `npm run change` will also produce a prompt via `chg`.

src/app/templates/scripts/_version.js
misteroneill: Yeah. This is run as the npm `"version"` script, at which point the `package.json` has been version-bumped, but it has not been committed yet.

docs/standards.md
gkatsev: Ah, cool. It should be documented that that is what happens. It isn't obvious to me.

src/app/templates/scripts/_version.js
gkatsev: :+1:

docs/standards.md
misteroneill: Makes sense. :+1:

gkatsev:
LGTM

PR #52, opened 2016-02-03
Budo

misteroneill:
This rewrites `scripts/server.js` to use `budo`, which is a more robust server that handles static file serving, watching, and live reload.

Generated projects will eliminate all the `"watch:*"` scripts and handle _all_ that behavior with the `"start"` script.

~~The way the watch-build cycle currently works is that it invokes `exec()` to call out to npm scripts. This is not the most efficient way of doing this, but ideally the spellbook will replace this in the long term. Alternatively, we can more build scripting to `scripts/` so that we don't have to do the easy, but slow thing.~~ This is now greatly improved by not calling out to npm all the time.

It also adds a `vjsgenclean` script, which will remove files and `package.json` fields that were added by the previous version of the generator, but are no longer needed by the new version.


gkatsev:
It would be nice to use budo's watchify for the main js stuff. `budo`'s `serve` property tells it what to serve the main javascript file as, so, you could call it `dist/videojs-foo.js` and then the `index.html` includes that (it already does). You would then give `budo` the path to the package's main file. I've tried just using `process.cwd()` but it didn't work for me in videojs: https://github.com/videojs/video.js/pull/2953/files

I'm ok with leaving this change out for now and fixing it up in a later release.

misteroneill:
I agree 100%, but budo seems to expect that its entry point(s) are not distinct from the output files. So, unless we change the `index.html` files to load `src/plugin.js` (which is an option), I don't think its built-in watchify would work. Also, the `serve` property only works if you're using its generated HTML.

That said, keeping a `browserify` object around in memory and bundling it whenever the watch triggers might be a good way to do things - almost certainly faster than `exec()`ing npm scripts!

gkatsev:
Nah, it works with a custom one as well. It'll either use the `index.html` in the root of where you run it from or you can give it a separate file. The `serve` option basically says that `{host}:{port}/{serve}` will return the output from watchify.

misteroneill:
Oh, cool, that seems counter to the docs, but they're a bit limited. I may experiment with that. :+1:

bin/videojs-generator-cleanup.js
```diff
@@ -0,0 +1,82 @@
...
+      'scripts.start:serve',
+      'scripts.watch',
+      'scripts.watch:css',
+      'scripts.watch:js',
+      'scripts.watch:test'
+    ]
+  }
+};
+
+var major = Number(require('../package.json').version.split('.')[0]);
```
incompl: Consider https://github.com/npm/node-semver

incompl:
LGTM

bin/videojs-generator-cleanup.js
misteroneill: Good call. :+1:

misteroneill:
Thanks @incompl! Any thoughts, @gkatsev?

src/app/templates/scripts/_server.js
```diff
@@ -1,26 +1,201 @@
...
+    let regex = new RegExp(keys[i]);
+
+    if (regex.test(file)) {
+      return handlers[keys[i]];
+    }
+  }
+};
+
+mkdirp('dist');
+bundle('js');
```
gkatsev: this is to make sure that we are working with the latest bundles when the server starts, right?

src/app/templates/scripts/_server.js
```diff
@@ -1,26 +1,201 @@
...
+  ])
+  .on('watch', (event, file) => {
+    const handler = findHandler(file);
+
+    console.log(`detected a "${event}" event in "${file}"`);
+
+    if (handler) {
+      handler(event, file);
+    } else {
+      console.log(`detected a "${event}" event in unmatched file "${file}"`);
```
gkatsev: this log is superfluous/duplicated of line 193. We should either move the above line into the if or remove this line.

src/app/templates/scripts/_server.js
```diff
@@ -1,26 +1,201 @@
...
+};
+
+/**
+ * A collection of functions which are mapped to strings that are used to
+ * generate RegExp objects. If a filepath matches the RegExp, the function
+ * will be used to handle that watched file.
+ *
+ * @type {Object}
+ */
+const handlers = {
```
gkatsev: sounds like handler should be a `Map`. Or at least, we should make it one in a future update that drops support for older nodes.

gkatsev:
One comment/change, one future enhancement.
LGTM.

src/app/templates/scripts/_server.js
misteroneill: Right.

src/app/templates/scripts/_server.js
misteroneill: Good point. I haven't ever seen that message come up since it handles everything it's watching!

src/app/templates/scripts/_server.js
misteroneill: Yeah, I agree.

gkatsev:
LGTM

PR #53, opened 2016-02-11
Spellbook - NOT READY

misteroneill:


misteroneill:
I'm gonna close this for now because I'm not sure if/when we'll get around to it (or whether we should).

gkatsev:
I still think we shou.

PR #66, opened 2016-05-05
Server Improvements

misteroneill:
- Throttle watch handler functions, so we don't trigger a bunch of builds and reloads when multiple files are changed at once.
- Make sure we bundle langs and Sass before starting server.
- For simplicity, bundle JS and tests when either changes.
- Do not watch `test/dist` file(s).

### DIFF
--- a/src/app/templates/scripts/_server.js
+++ b/src/app/templates/scripts/_server.js
@@ -43,7 +43,35 @@ const dests = {
   tests: 'test/dist/bundle.js'
 };
 
-const bundlers = {
+const tasks = {
+<% if (lang) { -%>
+
+  lang(resolve) {
+    vjslangs(srces.langs, dests.langs);
+    resolve();
+  },
+<% } -%>
+<% if (sass) { -%>
+
+  sass(resolve, reject) {
+    sass.render({
+      file: srces.css,
+      outputStyle: 'compressed'
+    }, (err, result) => {
+      if (err) {
+        reject(err.message);
+      } else {
+        fs.writeFile(dests.css, result.css, (err) => {
+          if (err) {
+            reject(err.message);
+          } else {
+            resolve();
+          }
+        });
+      }
+    });
+  },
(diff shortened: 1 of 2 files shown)


src/app/templates/scripts/_server.js
```diff
@@ -67,17 +68,29 @@ const bundlers = {
...
+        .on('finish', resolve)
+        .on('error', reject);
+    });
+  }
 };
 
 mkdirp.sync('dist');
 
+<% if (lang) { %>bundle('lang');<% } -%>
+<% if (sass) { %>bundle('sass');<% } -%>
```
incompl: Is it worth it to go through the bundle function for these, since they're special cased?

src/app/templates/scripts/_server.js
```diff
@@ -115,36 +128,23 @@ Promise.all([bundle('js'), bundle('tests')]).then(() => {
...
     /**
-     * Handler for JavaScript source.
+     * Handler for JavaScript source and tests.
      *
      * @param  {String} event
      * @param  {String} file
      */
-    '^src/.+\.js$'(event, file) {
-      console.log('re-bundling javascript and tests');
+    '^[st][re][cs]t?/.+\.js$': _.debounce((event, file) => {
```
incompl: How about `'^(src|test)'` instead?

misteroneill: Haha, yeah, that's better. 😄

src/app/templates/scripts/_server.js
misteroneill: Well, the reason is that they're called in two places and I like the idea of each of them being called through the same function. I have a better abstraction in mind, though.

incompl: Meant more as opposed to a separate function, but if you have something in mind, I'm sure it'll do the job too

incompl:
LGTM

PR #69, opened 2016-05-06
Only browserify-shim dist files

### LINKED ISSUE #58: Move browserify transforms to dependencies.

misteroneill:
Fixes #58 and #70 and issues with using generated plugins as a dependency of a project.

### DIFF
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -132,6 +132,9 @@ const packageJSON = (current, context) => {
         'build:js:uglify'
       ]),
 
+      // Babel is a run in a distinct step (vs. using babelify) because we want
+      // the transpiled code to be what's provided to module consumers using
+      // Node or Browserify.
       'build:js:babel': 'babel src -d es5',
 
       'build:js:bannerize': scriptify([
@@ -139,8 +142,11 @@ const packageJSON = (current, context) => {
         '--banner=scripts/banner.ejs'
       ]),
 
+      // The browserify-shim transform is included ONLY in the build step as
+      // we do not want consuming projects to receive a shimmed module - if we
+      // shim always, we cause obscure errors!
       'build:js:browserify': scriptify([
-        'browserify . -t browserify-shim -t browserify-versionify -s %s -o dist/%s.js'
+        'browserify . -t browserify-shim -s %s -o dist/%s.js'
       ]),
 
       'build:js:uglify': scriptify([
@@ -168,9 +174,20 @@ const packageJSON = (current, context) => {
     'author': context.author,
     'license': context.licenseName,
 
+    'browserify': {
+
+      // Unlike browserify-shim, we want to apply the browserify-versionify
+      // shim ALWAYS because we want the version of this project to always be
+      // available to consumers.
+      transform: ['browserify-versionify']
+    },
+
     'browserify-shim': {
       'qunit': 'global:QUnit',
       'sinon': 'global:sinon',
+
+      // video.js is shimmed for the distributable because we don't want to
+      // build it into every plugin's distributed files.
       'video.js': 'global:videojs'
     },
 
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -140,7 +140,7 @@ const packageJSON = (current, context) => {
       ]),
 
       'build:js:browserify': scriptify([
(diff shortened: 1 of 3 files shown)


src/app/package-json.js
```diff
@@ -168,13 +168,6 @@ const packageJSON = (current, context) => {
     'author': context.author,
     'license': context.licenseName,
 
-    'browserify': {
-      transform: [
-        'browserify-shim',
-        'browserify-versionify'
```
gkatsev: we want to keep `browserify-versionify`, no?

misteroneill: I guess. Will browserify merge the `-t` transforms with those in the `package.json`?

gkatsev: Hm... I'm actually not sure how it handles it. Would be good to try it out.
Also, we'll want to make sure that `browserify-versionify` is set as a dependency and not a dev dependency.

misteroneill: Maybe we should wait on merging until we have a deeper understanding.

brookslyrette: Just a note. I added this on a project and found there is a small bug in the build script. 

Line 143 should read: 

`'browserify . -t browserify-shim -t browserify-versionify -s %s -o dist/%s.js'`

misteroneill: Thanks!

gkatsev: Looks like both get included, though I don't know about ordering guarantees yet.
Also, if it's being configured through both the cli and the config it may be run twice.

misteroneill: I don't think ordering should matter in this case, but it's something to consider. I'll change this back to do versionify always and shim only for dist.

src/app/templates/scripts/_build-test.js
```diff
@@ -10,6 +10,8 @@ glob('test/**/*.test.js', (err, files) => {
   }
   browserify(files)
     .transform('babelify')
+    .transform('browserify-shim')
+    .transform('browserify-versionify')
```
BrandonOCasey: I think we should only have to do `browserify-shim` here as the package.json entry should handle the `browserify-versionify` for us.

BrandonOCasey:
I think `browserify-versionify` needs to be a dependency so that other objects can respect the global transform.

BrandonOCasey:
Going to do some testing to verify that this fixes the issues with one project using another, LGTM though

src/app/package-json.js
```diff
@@ -132,15 +132,21 @@ const packageJSON = (current, context) => {
...
         'bannerize dist/%s.js',
         '--banner=scripts/banner.ejs'
       ]),
 
+      // The browserify-shim transform is included ONLY in the build step as
+      // we do not want consuming projects to receive a shimmed module - if we
+      // shim always, we cause obscure errors!
       'build:js:browserify': scriptify([
-        'browserify . -s %s -o dist/%s.js'
+        'browserify . -t browserify-shim -s %s -o dist/%s.js'
```
BrandonOCasey: this needs to be changes to a global transform: `-g browserify-shim` rather than `-t browserify-shim`

src/app/templates/scripts/_build-test.js
```diff
@@ -10,6 +10,7 @@ glob('test/**/*.test.js', (err, files) => {
   }
   browserify(files)
     .transform('babelify')
+    .transform('browserify-shim')
```
BrandonOCasey: also needs to be global: `.transform('browserify-shim', {global: true})`

src/app/templates/scripts/_server.js
```diff
@@ -79,8 +79,7 @@ const tasks = {
     standalone: nameify('%s'),
     transform: [
       'babelify',
-      'browserify-shim',
-      'browserify-versionify'
+      'browserify-shim'
```
BrandonOCasey: global again `['browserify-shim', {global: true})]`

src/app/templates/scripts/_server.js
```diff
@@ -89,8 +88,7 @@ const tasks = {
     entries: srces.tests,
     transform: [
       'babelify',
-      'browserify-shim',
-      'browserify-versionify'
+      'browserify-shim'
```
BrandonOCasey: final global

BrandonOCasey:
LGTM!

PR #75, opened 2016-06-22
remove direct specification of karma plugins

BrandonOCasey:
As per the [documentation](https://karma-runner.github.io/0.13/config/plugins.html)
> By default, Karma loads all sibling NPM modules which have a name starting with karma-*.

So I have removed all references to plugins from the karma config as well as making the logic for wether to use detectBrowser simpler by using the enabled option.

### DIFF
--- a/src/app/templates/test/_karma.conf.js
+++ b/src/app/templates/test/_karma.conf.js
@@ -1,34 +1,23 @@
 module.exports = function(config) {
-  var browsers = config.browsers;
-  var frameworks = ['qunit'];
-  var plugins = ['karma-qunit'];
-
-  var addBrowserLauncher = function(browser) {
-    plugins.push('karma-' + browser.toLowerCase() + '-launcher');
+  var detectBrowsers = {
+    enabled: false,
+    usePhantomJS: false
   };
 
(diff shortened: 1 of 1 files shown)


src/app/templates/test/_karma.conf.js
```diff
@@ -1,34 +1,23 @@
 module.exports = function(config) {
-  var browsers = config.browsers;
-  var frameworks = ['qunit'];
-  var plugins = ['karma-qunit'];
-
-  var addBrowserLauncher = function(browser) {
-    plugins.push('karma-' + browser.toLowerCase() + '-launcher');
+  var detectBrowsers = {
+    enabled: false,
```
gkatsev: This disables the `detectBrowsers` plugin?

gkatsev:
LGTM

src/app/templates/test/_karma.conf.js
BrandonOCasey: yeah it my testing it does and it also says so in the [README](https://github.com/litixsoft/karma-detect-browsers)

misteroneill:
As long as this doesn't break the `test:foo` npm scripts... LGTM

PR #76, opened 2016-07-07
Update .gitignore

chivio3:


gkatsev:
This seems to be changing the gitignore to support visual studio c++, ghc, and fortran languages. Those are completely unrelated from what this project deals with, so, we cannot accept this patch.

PR #80, opened 2016-08-20
Make ghooks have options (including off)

### LINKED ISSUE #73: Multiple ghooks options
Not everyone wants to `"npm test"` for every push - especially when it launches a bunch of browser processes. Provide an alternative option: `"npm run lint"`.

misteroneill:
It doesn't make a ton of sense for "test" to be the default and only option for ghooks. This adds "lint" as another option (the default as well) and allows the user to choose "none".

Fixes #73

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -304,8 +304,8 @@ export default yeoman.generators.Base.extend({
     this._currentPkgJSON = this.fs.readJSON(this.destinationPath('package.json'), null);
 
     this._ghooksOptions = {
-      lint: 'Check code quality using videojs-standard',
-      test: 'Run unit tests using Karma',
+      lint: 'Check code quality',
+      test: 'Check code quality and run tests',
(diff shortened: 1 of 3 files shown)


src/app/index.js
```diff
@@ -294,6 +303,12 @@ export default yeoman.generators.Base.extend({
 
     this._currentPkgJSON = this.fs.readJSON(this.destinationPath('package.json'), null);
 
+    this._ghooksOptions = {
+      lint: 'Check code quality using videojs-standard',
+      test: 'Run unit tests using Karma',
```
BrandonOCasey: We may want specify that test will also lint due to the `pretest` command in `package.json`

misteroneill: Sounds good. And maybe leave out the "using Karma" bit as that's kind of redundant information.

BrandonOCasey:
LGTM

PR #81, opened 2016-08-20
Don't lint scripts

### LINKED ISSUE #77: scripts/server.js does not pass linter when Sass is enabled
[Here](https://github.com/videojs/generator-videojs-plugin/blob/master/src/app/templates/scripts/_server.js#L64), causes `"err" is already declared in the upper scope.`

misteroneill:
There have been too many issues with generated code failing linting; so, rather than a bunch of patches to fix virtually pointless problems, let's just not lint the `scripts/` directory.

Fixes #77

### DIFF
--- a/src/app/templates/scripts/_server.js
+++ b/src/app/templates/scripts/_server.js
@@ -54,6 +54,8 @@ const tasks = {
 <% if (sass) { -%>
 
   sass(resolve, reject) {
+    const finish = (err) => err ? reject(err.message) : resolve();
+
     sass.render({
       file: srces.css,
       outputStyle: 'compressed'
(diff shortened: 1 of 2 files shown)


misteroneill:
Also fixes a linter problem because might as well. And it's sort of ironic. 😄

BrandonOCasey:
LGTM

src/app/templates/scripts/_server.js
```diff
@@ -61,9 +61,9 @@ const tasks = {
       if (err) {
         reject(err.message);
       } else {
-        fs.writeFile(dests.css, result.css, (err) => {
-          if (err) {
-            reject(err.message);
+        fs.writeFile(dests.css, result.css, (errr) => {
```
BrandonOCasey: make one of these `error` and the other `err` rather than using `errr` as it almost seems like a miss-spelling

misteroneill: That sounds reasonable. :smile:

gkatsev: the other option is to just err all the way down.

misteroneill: The linter complains about that because it's overriding the `err` from a previous scope.

gkatsev: *shadowing*, though. I guess either way is fine.

BrandonOCasey:
LGTM

PR #83, opened 2016-09-07
add bundle collapser to the generator

BrandonOCasey:


src/app/package-json.js
```diff
@@ -143,8 +145,13 @@ const packageJSON = (current, context) => {
...
       ]),
 
+      'build:js:collapse': scriptify([
+        'bundle-collapser dist/%s.js',
+        '-o dist/%s.min.js'
+      ]),
+
       'build:js:uglify': scriptify([
-        'uglifyjs dist/%s.js',
+        'uglifyjs dist/%s.min.js',
```
misteroneill: Should this suffix be changed to `coll` or something instead of `min`? This confused me momentarily.

BrandonOCasey: well I am just treating bundle collapsing as a minification step

misteroneill: Ok, cool, that's fine. What about overwriting `dist/%s.js`?

misteroneill:
LGTM

PR #86, opened 2016-09-09
update to babel 6

BrandonOCasey:
* re-generated README table of contents
* updated babel to babel 6

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -329,7 +329,8 @@ export default yeoman.generators.Base.extend({
       '_.editorconfig',
       '_.gitignore',
       '_.npmignore',
-      '_jsdoc.json'
+      '_jsdoc.json',
+      '_.babelrc'
     ];
 
     this._templatesToCopy = [
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -145,7 +145,8 @@ export default yeoman.generators.Base.extend({
       ghooks: configs.hasOwnProperty('ghooks') ? !!configs.ghooks : 'lint',
       lang: configs.hasOwnProperty('lang') ? !!configs.lang : false,
       license: this._licenseDefault,
-      sass: configs.hasOwnProperty('sass') ? configs.sass : false
+      sass: configs.hasOwnProperty('sass') ? configs.sass : false,
+      ie8: configs.hasOwnProperty('ie8') ? configs.ie8 : false
     };
 
     ['author', 'license', 'name', 'description'].forEach(key => {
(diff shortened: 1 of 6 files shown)


misteroneill:
Nice, I'll tinker with this over the weekend or early next week!

src/app/templates/_.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["es2015-loose"],
```
gkatsev: we probably should default to non-loose and maybe have an option for loose mode? Plugins that don't care or need IE8 don't want loose mode. Also, for IE8 support, we want the other two or three transforms from videojs.

misteroneill: Perhaps a "Do you need IE8 support?" prompt is in order...

misteroneill: It also appears that [babel-preset-es2015-loose](https://www.npmjs.com/package/babel-preset-es2015-loose) is deprecated and now built into [babel-preset-es2015](https://www.npmjs.com/package/babel-preset-es2015) as an option - `{loose: true}`.

BrandonOCasey: nice call, must have just come out

BrandonOCasey:
@Pat @gkatsev added the "support ie8 option" and tests for it

src/test/app.test.js
```diff
@@ -4,9 +4,9 @@ import _ from 'lodash';
 import fs from 'fs-extra';
 import path from 'path';
 import {assert, test as helpers} from 'yeoman-generator';
-
 import * as libs from './libs';
 import packageJSON from '../generators/app/package-json';
+import {assert as chaiAssert} from 'chai';
```
misteroneill: Is there a reason chai is needed here vs using the assert module provided by yeoman? It extends Node's [assert](https://nodejs.org/api/assert.html) module; so, you should have the methods you need available.

src/app/index.js
```diff
@@ -228,6 +229,11 @@ export default yeoman.generators.Base.extend({
       default: defaults.sass
     }, {
       type: 'confirm',
+      name: 'ie8',
+      message: 'Do you want to support ie8?',
```
misteroneill: Minor point, but I think "ie" should be capitalized here.

src/app/templates/_.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [["es2015", {"loose": true}]],
+<% if (ie8) { -%>
+	"plugins": [
```
misteroneill: Indenting looks off in this block, maybe?

src/app/index.js
BrandonOCasey: Yeah, may even want to spell it out just in case good call

src/test/app.test.js
BrandonOCasey: I just looked at the yeoman api docs and didn't see anything, but I never realized that it inherits from node assert. I will look into it

misteroneill:
LGTM

PR #87, opened 2016-09-09
Add rollup to the generator (NOT READY FOR MERGE)

BrandonOCasey:
* do we want to have `dist/{test,npm,browser}`? or do we want to serve a single concatenated file from rollup on npm? IMO we want to have an npm dir so that the user can require any module they want in the project without jumping through hoops. 
 based on #86


BrandonOCasey:
this is going to be replaced by spellbook

PR #92, opened 2016-09-24
Fix a possible delete property error

misteroneill:
This change fixes a potential issue when generating a new project and selecting "none" for the ghooks option. It will cause an error to be thrown because the key does not exist. This causes the project to fail to generate.


BrandonOCasey:
maybe add a unit test for this option too?

BrandonOCasey:
LGTM

PR #94, opened 2016-09-24
Do not update CHANGELOG for prereleases

### LINKED ISSUE #72: CHANGELOG should not update on prereleases

misteroneill:
Fixes #72

### DIFF
--- a/src/app/package-json.js
+++ b/src/app/package-json.js
@@ -32,7 +32,7 @@ const DEFAULTS = {
     'karma-safari-launcher': '^0.1.1',
     'lodash': '^4.11.2',
     'mkdirp': '^0.5.1',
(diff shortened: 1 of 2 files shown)


src/app/package-json.js
```diff
@@ -33,6 +32,7 @@ const DEFAULTS = {
     'karma-safari-launcher': '^0.1.1',
     'lodash': '^4.11.2',
     'mkdirp': '^0.5.1',
+    'node-semver': '^5.3.0',
```
BrandonOCasey: I think this should be called `semver`. I know that it works (as I did the same thing in spellbook) but it seems like all of the examplse on `https://github.com/npm/node-semver` use `semver` and not `node-semver`. It's also less confusing when we go to require the package.

PR #95, opened 2016-10-06
update vjs-standard to 5.0.0

BrandonOCasey:


misteroneill:
@BrandonOCasey This is great, but it needs to be rebased and then the linting should pass. It's mostly `let` to `const` changes, though. 😄 

It'd be nice to get this in a minor update.

misteroneill:
LGTM

Issue #88, opened 2016-09-13
Add tests for browserify, node require, webpack, and rollup use


BrandonOCasey:
Similar to what we have in video.js, but makes it easier for users to support all of the package bundlers

Issue #84, opened 2016-09-09
rename main file from plugin.js to index.js


BrandonOCasey:
plugin.js kind of makes sense, but index.js makes it way easier to work with the main source file of a plugin by only having to refer to the top level directory, and index.js is used much more often for node.js projects (it's usually index.js or the projects name as the main file)

Example

``` JavaScript
// if the file is `src/index.js' we can now require the plugin by referencing 
require('src');
```

misteroneill:
I'm fine with renaming `src/plugin.js` to `src/index.js`.

That said, `"main"` should still point to `es5/index.js` or `lib/index.js` (it might make sense to rename it) because it should point to the Babel output, so others can use the plugin via Node/Browserify without needing Babel.

BrandonOCasey:
yeah I agree it should point to `es5/index.js` rather than `src/index.js` since we want to point to the es5 version

Issue #74, opened 2016-06-22
Load test dependancies from one location


BrandonOCasey:
## Problem

Right now it's very awkward to edit tests dependancies as they are stored in two places, it would be really nice to load them all from the same place to prevent errors in one or the other. (test/index.html vs karma config)
## Solutions
1. Use karma's builtin server/watch/debug functionality on `start` so that we don't have to maintain test builds
2. Inject dependancies in the karma config into the `test/index.html` file using another build step
## Choice

I think that we should really investigate number one as it will make this whole project simpler if we can get it to work and we won't have to maintain any test html or test build ourselves.
## Implementation
- proxy karma's debug page to the current `:9999/test` route so that everything is seemless
- use https://github.com/karma-runner/karma-qunit/pull/41 to show the qunit UI when it is merged
- configure `npm run start` and other applicable commands to do `karma start test/karma.conf.js --no-single-run`
- change the karma configuration file so that it allows command line overrides IE `singleRun: config.singleRun || false`
- configure karma to run on watch with `autoWatch` set to true, so we can auto re-run tests

misteroneill:
Sounds like a great idea - I'm all for it.

gkatsev:
I think I would prefer doing the reverse of that if anything. Generate the test file separately use that bundle in `test/index.html` and karma.
Also, this ties in with https://github.com/videojs/video.js/issues/3543, where maybe the place where we put the test files isn't the ideal location for how we build and test our code.

misteroneill:
If I'm understanding correctly, that's kinda what it does now: we have `test/dist/bundle.js` that includes the tests and source in one big package.

I think @dmlap didn't like that because we're not testing the `dist/videojs-foo.js` file, rather a compilation specific to tests, which seems like a fair point to me.

BrandonOCasey:
I think that we still want to do this so that we don't have a silly `test/index.html` which has to copy all of the files from `test/karma.conf.js` verbatim

misteroneill:
What about those of us who like to load their tests in the browser without all that Karma nonsense!? 😉

BrandonOCasey:
haha well you won't even need to know about karma

PR #96, opened 2016-10-11
integrate with spellbook

### LINKED ISSUE #90: include scripts via a separate package, included with generator
- remove scripts from the generator and make a separate package for them
- have that package get installed via the generator by adding it as a dev dependency
- have `package.json` run scripts from that package

BrandonOCasey:
This will fix #90, #88, #85, #84, #74, and #59


misteroneill:
This fixes #90

package.json
```diff
@@ -42,7 +42,6 @@
   ],
   "dependencies": {
```
misteroneill: I think we're missing the spellbook dependency.

src/app/index.js
```diff
@@ -140,15 +139,26 @@ export default yeoman.generators.Base.extend({
...
     let defaults = {
       bower: configs.hasOwnProperty('bower') ? !!configs.bower : true,
-      changelog: configs.hasOwnProperty('changelog') ? !!configs.changelog : true,
-      docs: configs.hasOwnProperty('docs') ? !!configs.docs : false,
       ghooks: configs.hasOwnProperty('ghooks') ? !!configs.ghooks : 'lint',
-      lang: configs.hasOwnProperty('lang') ? !!configs.lang : false,
       license: this._licenseDefault,
-      sass: configs.hasOwnProperty('sass') ? configs.sass : false,
+
+      noShimVideojs: configs.hasOwnProperty('noShimVideojs') ? !!configs.noShimVideojs : false,
```
misteroneill: I'm curious why we need this new option.

src/app/index.js
```diff
@@ -618,6 +602,8 @@ export default yeoman.generators.Base.extend({
     if (this.options.hurry) {
       return;
     }
+    shelljs.exec('git add --all');
+    shelljs.exec('git commit -a -m "generator run"');
```
misteroneill: I'm not sure about this. I don't really like the idea of creating commits implicitly.

src/app/index.js
```diff
@@ -382,15 +386,8 @@ export default yeoman.generators.Base.extend({
         templates: []
       },
       scripts: {
```
misteroneill: Can this object be removed entirely?

src/app/index.js
BrandonOCasey: should we just do `git init` and leave it at that?

misteroneill: I think so. It's kind of a fine line between convenience and imposition and I feel like this crosses it just slightly.

For example, when running against an existing project, we don't want to be creating commits with files that may need manual cleanup.

package.json
BrandonOCasey: The generator itself will not have to depend on `spellbook` unless we switch the build system to use it. The only problem with that would be the lack of support for running mocha tests, and even if we wanted to switch to qunit there is no good way to run qunit tests in node.

misteroneill: Yeah. I wasn't thinking! 😆

src/app/index.js
misteroneill: Yeah, seems a bit too fiddly and I worry that users will get themselves into trouble.

package.json
```diff
@@ -42,7 +42,6 @@
...
     "chalk": "^1.0.0",
-    "chg": "^0.3.2",
```
misteroneill: I actually need this in `master` now for an official CHANGELOG. Can this change be undone?

misteroneill:
As discussed, we won't be supporting Windows for the immediate release of v3.0.0 but an eventual downstream release of videojs-spellbook should fix that without the need for another generator release.

BrandonOCasey:
@misteroneill Ready for re-review :)

src/app/templates/_.gitignore
```diff
@@ -30,6 +30,4 @@ node_modules/
 
 # Build-related directories
 dist/
-docs/api/
-es5/
-test/dist/
+
```
misteroneill: I feel like these should be ignored by Git...

BrandonOCasey: well `test/dist`, `docs/api` and `es5` will no longer exist in the root folder. All of them will be in `dist/test` or `dist/docs` or `dist/es5`

misteroneill: Ohhhh, right. Carry on!

misteroneill:
A couple issues I noticed in testing:

- `localhost:9999/test/` no longer runs QUnit tests, which is fine, but the link from `index.html` in the root should be fixed to point, I guess, to `localhost:9876`.

BrandonOCasey:
@misteroneill 
* The issues that you are seeing with: `localhost:9999/test/` may be due to uglify, running it locally with a test project causes the tests to work for me. What command where you running to start your server?
* Do we just want to cut out the scope at the front of the npm package?
* The css/sass failures appear to both be related to the banner, I will see if I can find out why.
* I will fix sb-release so that it can be run with npm version, npm run version, or as sb-release

misteroneill:
- I started the server with `npm start`. Then `localhost:9999/test/` showed `NOT FOUND`. I don't recall if anything was logged in the console.
- The scope should just be in the package name. npm will install it in `@foo/bar` but all we have to do is put it in the name. Spellbook should probably "descope" project names if it gets it from `package.json`.
- Sounds good on `sb-release`! 👍

misteroneill:
I changed the base branch because v3 will live on the `next` branch for a while as we maintain support for v2.

BrandonOCasey:
ahh yeah banner is killing the build for you as you have a single quote in your name hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm. The one setback of the command line shows its ugly head. let me see what I can do about it.

BrandonOCasey:
@misteroneill Ready to be tested again!

misteroneill:
Seems to be working great. Thanks @BrandonOCasey!

PR #97, opened 2016-10-12
Add generator-version

### LINKED ISSUE #91: Store generator version in package.json
This would be really useful to know what version of the generator was used most recently in a project.

misteroneill:
Fixes #91

### DIFF
--- a/src/app/generator-version.js
+++ b/src/app/generator-version.js
@@ -1,0 +1,9 @@
+
+/**
+ * Gets the version of the generator.
+ *
+ * @return {String}
+ */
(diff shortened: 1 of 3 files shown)


BrandonOCasey:
can this be done in a file that already exists, like .yo-rc?

misteroneill:
Yes, but .yo-rc isn't (and shouldn't be) checked into package.json. We can use this metadata for scripts that need to detect the version of the generator used to build a project - for example, a 2.x to 3.x migration script.

BrandonOCasey:
ok LGTM then

PR #100, opened 2016-11-04
videojs-generator-cleanup Updates for v3

misteroneill:
Tested and working.


bin/videojs-generator-cleanup.js
```diff
@@ -34,51 +43,140 @@ var configs = {
...
     ]
+  },
+  v3: {
+    removals: [
+      '.babelrc',
+      'dist',
+      'docs/api',
+      'es5',
+      'jsdoc.json',
+      'scripts',
```
BrandonOCasey: we may want to specify specific scripts here, just in case they are using scripts of their own in this directory.

bin/videojs-generator-cleanup.js
```diff
@@ -34,51 +43,144 @@ var configs = {
...
+    },
+    pkg: [
+      'browserify',
+      'dependencies.browserify-versionify',
+      'dependencies.video.js',
+      'devDependencies.babel-cli',
+      'devDependencies.babel-plugin-transform-es3-member-expression-literals',
+      'devDependencies.babel-plugin-transform-es3-property-literals',
+      'devDependencies.babel-plugin-transform-object-assign',
+      'devDependencies.babel-preset-es2015',
```
BrandonOCasey: may want to include `babel-preset-es2015-loose`, I'm not sure if we ever had it in a patch (we may have), but it was deprecated right after and should no longer be needed.

bin/videojs-generator-cleanup.js
```diff
@@ -34,51 +43,144 @@ var configs = {
...
+      'scripts.build:css:sass',
+      'scripts.build:js',
+      'scripts.build:js:babel',
+      'scripts.build:js:bannerize',
+      'scripts.build:js:browserify',
+      'scripts.build:js:collapse',
+      'scripts.build:js:uglify',
+      'scripts.build:lang',
+      'scripts.build:test',
+      'scripts.docs',
```
BrandonOCasey: `start:serve`, `version`, `watch:docs`, `watch:babel`, `watch:browserify` `watch:test`. I took these from

bin/videojs-generator-cleanup.js
gkatsev: We shouldn't need the `-loose` preset. `loose` is available as an option on the es2015 preset. So, we just need to enable the option whenever we would use `es2015-loose`.

BrandonOCasey: yeah but this list is for removing it from dependencies

BrandonOCasey:
Ok just finished testing on HLS (I think they are sill on v1). Here are some things I noticed:
* babel 5.8.0 is still in devDependencies (could be a leftover from v1 -> v2)
* watch-test script is still in scripts (could be a leftover from v1 -> v2)
* shelljs is still a devDependency 

**Spellbook**: (this is for me later)
* I may need to support `src` folder as the folder where just js resides
* A small change on how I find the main file, currently expecting it to be index.js, but we could just use the jsnext:main from package.json.

misteroneill:
> re-running the generator does some scary things, and if I didn't know how the generator worked intimately I don't know if I would want to run it. ...

Unfortunately, Yeoman does not (as far as I can tell) offer the capability to prevent user choice when it comes to file conflicts, which is why we have this [Recommendations](https://github.com/videojs/generator-videojs-plugin#recommendations) section of the README.

I'll look into the rest tomorrow.

misteroneill:
> PULL_REQUEST_TEMPLATE.md & ISSUE_TEMPLATE.md should be moved to .github

We can do this another time.

> babel 5.8.0 is still in devDependencies (could be a leftover from v1 -> v2)

This was changed in the 1.x line. I'll have it removed from the 1->2 cleanup script.

> watch-test script is still in scripts (could be a leftover from v1 -> v2)

HLS has a bunch of custom scripts. We can't/shouldn't try to delete scripts we didn't create. Unused watch scripts are removed in 1->2.

> lint, clean, build, start, and test still exists after cleaning (not sure if this will be overwritten when I update it)

I'm not sure what this is referring to. Those scripts are still part of the generator - we don't want to remove them... If you're running the cleanup script before applying v3, then, you shouldn't. But I will make that more clear in documentation. :)

> Should we make preversion & prerelease do test, followed by lint?

Does spellbook not lint as part of the test script? Regardless, I would say lint first because it's so much quicker and lower impact.

> shelljs is still a devDependency

I think this was added to HLS and is another thing we can't/shouldn't try to delete it.

> files in package.json is not correct. es5/, scripts/, and index.html need to be removed.
> browserify-shim needs to be removed (not sure if it will be overridden)

I believe this is fixed by running the generator.

> video.js and global can be removed from dependencies/devDependencies if they have the same version range as videojs-spellbook. IE: "global": "^4.3.1", & "video.js": "^5.11.7",

Hmm, gonna have to think about this one.

> test/karma still exists

Yes, it is removed by the 1->2 process, though. Maybe it should run _all_ cleanup scripts for prior versions too (in order)? Or maybe not. This isn't meant to be a kitchen-sink sort of thing.

> it has a change script? do we still have this? I think this may have been lost

Adding it to removals.

> es5 needs to be removed from files in package.json

This is handled by running the generator.

> babel: 5.8.35 is still a devDependency but spellbook will handle this
> chg is still a devDependency but spellbook will include it for us

Adding both to removals.

misteroneill:
Also, in order to properly use the cleanup script, the generator's `version` needs to be manually changed to 3 so that it thinks it's updating to 3.

I just pushed a change that will run _all_ major versions up to the current one.

BrandonOCasey:
@misteroneill will test again today

BrandonOCasey:
Seems good now, also spellbook does not run lint as part of test. Should I implement that?

misteroneill:
I think it makes sense.

PR #101, opened 2016-11-04
videojs-plugin-check Updates for v3

misteroneill:
Tested and working.

Also, this includes the commit from #100 - so, only review changes to files in `src/check/`.

### DIFF
--- a/src/check/package-json.test.js
+++ b/src/check/package-json.test.js
@@ -52,8 +52,8 @@ tap.ok(
 
 tap.ok(_.isPlainObject(pkg.spellbook), 'package.json has "spellbook" object');
 
-spellbookConfigs.forEach(sbc => {
-  tap.equal(typeof sbc, 'boolean', `package.json "spellbook.${sbc}" is a boolean`);
+Object.keys(pkg.spellbook).forEach(k => {
+  tap.equal(typeof pkg.spellbook[k], 'boolean', `package.json "spellbook.${k}" is a boolean`);
 });
 
 ['videojs', 'videojs-plugin'].forEach(kw => {
@@ -69,4 +69,4 @@ coreScripts.forEach(script => {
   );
 });
 
-tap.strictSame(pkg.files, publishedFiles, 'package.json "files" has expected contents');
+tap.ok(publishedFiles.every(f => pkg.files.indexOf(f) > -1), 'package.json "files" has expected contents');
--- a/src/check/package-json.test.js
+++ b/src/check/package-json.test.js
@@ -50,8 +50,8 @@ tap.ok(
 
 tap.ok(_.isPlainObject(pkg.spellbook), 'package.json has "spellbook" object');
 
-Object.keys(pkg.spellbook).forEach(k => {
-  tap.equal(typeof pkg.spellbook[k], 'boolean', `package.json "spellbook.${k}" is a boolean`);
+_.forEach(pkg.spellbook, (value) => {
+  tap.ok(_.isBoolean(value) || _.isObject(value), `package.json "spellbook.${k}" is an appropriate type`);
 });
 
 ['videojs', 'videojs-plugin'].forEach(kw => {
--- a/src/check/package-json.test.js
+++ b/src/check/package-json.test.js
@@ -8,8 +8,25 @@ const coreScripts = [
   'build',
   'clean',
   'lint',
+  'release',
   'start',
-  'test'
+  'test',
+  'watch'
+];
+
+const spellbookConfigs = ['js', 'css', 'lang', 'docs'];
+
+const publishedFiles = [
+  'CONTRIBUTING.md',
+  'CHANGELOG.md',
+  'README.md',
+  'dist/docs',
+  'dist/lang',
+  'dist/es5',
+  'dist/browser',
+  'index.html',
+  'src/',
+  'docs/'
 ];
 
 tap.ok(_.isPlainObject(pkg), 'package.json exists');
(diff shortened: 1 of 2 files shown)


src/check/structure.test.js
```diff
@@ -1,11 +1,14 @@
...
 
-tap.ok(isDir('scripts'), 'the "scripts/" directory exists');
 tap.ok(isDir('src'), 'the "src/" directory exists');
-tap.ok(isFile('src/plugin.js'), 'the "src/plugin.js" file exists');
+tap.ok(isDir('src/js/'), 'the "src/js/" directory exists');
+tap.ok(isFile('src/js/index.js'), 'the "src/plugin.js" file exists');
 tap.ok(isDir('test'), 'the "test/" directory exists');
-tap.ok(isFile('test/karma.conf.js'), 'the "test/karma.conf.js" file exists');
-tap.ok(isFile('test/plugin.test.js'), 'the "test/plugin.test.js" file exists');
+tap.ok(isFile('test/index.test.js'), 'the "test/plugin.test.js" file exists');
```
BrandonOCasey: description needs to be updated here

BrandonOCasey: description needs to be updated here

src/check/structure.test.js
```diff
@@ -1,11 +1,14 @@
 import {isDir, isFile} from './lib';
 import tap from 'tap';
 
-tap.ok(isDir('scripts'), 'the "scripts/" directory exists');
 tap.ok(isDir('src'), 'the "src/" directory exists');
-tap.ok(isFile('src/plugin.js'), 'the "src/plugin.js" file exists');
+tap.ok(isDir('src/js/'), 'the "src/js/" directory exists');
+tap.ok(isFile('src/js/index.js'), 'the "src/plugin.js" file exists');
```
BrandonOCasey: description needs to be updated here

BrandonOCasey: seems like there should be an array of `files` and `dirs` to check, that way we won't have an issue with diverging descriptions.

src/check/package-json.test.js
```diff
@@ -28,7 +25,7 @@ tap.ok(
   'package.json has "author" property'
 );
 
-tap.equal(pkg.main, 'es5/plugin.js', 'package.json "main" is "es5/plugin.js"');
+tap.equal(pkg.main, 'dist/es5/index.js', 'package.json "main" is "dist/es5/index.js"');
```
BrandonOCasey: we also need to check for:
* `jsnext:main` here, as it is required by spellbook
* `generator-videojs-plugin.version` 
* check the files array (to make sure everything is published, but only what should be published.)
* is there any way to check the `.yo.rc` to verify that the `spellbook` configuration in `package.json` matches here?
* such as ie8, and css off
* should we check ghooks here against the `.yo-rc`

src/check/package-json.test.js
```diff
@@ -6,13 +6,10 @@ const pkg = require(`${process.cwd()}/package.json`);
...
 const coreScripts = [
   'build',
-  'build:js',
-  'build:test',
   'clean',
   'lint',
   'start',
-  'test',
-  'version'
+  'test'
```
BrandonOCasey: I think we need `watch`, `postversion`, `prepublish`, and `release`

src/check/structure.test.js
```diff
@@ -1,11 +1,14 @@
...
 tap.ok(isDir('test'), 'the "test/" directory exists');
-tap.ok(isFile('test/karma.conf.js'), 'the "test/karma.conf.js" file exists');
-tap.ok(isFile('test/plugin.test.js'), 'the "test/plugin.test.js" file exists');
+tap.ok(isFile('test/index.test.js'), 'the "test/plugin.test.js" file exists');
+tap.ok(isFile('.editorconfig'), 'the ".editorconfig" file exists');
+tap.ok(isFile('.gitignore'), 'the ".gitignore" file exists');
+tap.ok(isFile('.npmignore'), 'the ".npmignore" file exists');
+tap.ok(isFile('CHANGELOG.md'), 'the "CHANGELOG.md" file exists');
 tap.ok(isFile('index.html'), 'the "index.html" file exists');
 tap.ok(isFile('README.md'), 'the "README.md" file exists');
```
BrandonOCasey: We should also check:
* to make sure `src/css` contains an `index.css`, `index.scss`, or an `index.sass`. if css is on in the .yo-rc.
* a lang folder exists with json files if .yo-rc says so
* a docs folder with md files if .yo-rc says so

misteroneill: I can see a case for checking for an `index.[css|scss|sass]` file, but not the others. This isn't meant to be a comprehensive tool: it's meant to see if a directory meets the minimum criteria of required files/folders. Also, it can't check against `.yo-rc.json` because that file is ignored.

src/check/package-json.test.js
misteroneill: I'm good with `watch` becoming a required script, but I don't think version/prepublish scripts should be included even though the generator adds them.

src/check/package-json.test.js
misteroneill: - `jsnext:main` - Adding.
- `generator-videojs-plugin.version` - Adding.
- `files` - Adding.
- The rest is either related to optional features or relies on the `.yo-rc` which I want to avoid. But I will add verification of the spellbook configuration.

src/check/structure.test.js
misteroneill: On second thought, we can check the spellbook configuration in lieu of `.yo-rc.json`.

src/check/package-json.test.js
```diff
@@ -28,7 +40,19 @@ tap.ok(
...
+
+tap.ok(
+  isNonEmptyString(pkg['generator-videojs-plugin'].version),
+  'package.json "generator-videojs-plugin.version" exists'
+);
+
+tap.ok(_.isPlainObject(pkg.spellbook), 'package.json has "spellbook" object');
+
+Object.keys(pkg.spellbook).forEach(k => {
+  tap.equal(typeof pkg.spellbook[k], 'boolean', `package.json "spellbook.${k}" is a boolean`);
```
BrandonOCasey: spellbook config entries can be booleans (in the case where the option is turned off), but they can also be objects

misteroneill: How's that? It only looks like they're ever [set to booleans](https://github.com/videojs/generator-videojs-plugin/blob/master/src/app/package-json.js#L130-L139)...

BrandonOCasey: * `src`, `js`, `css`, `test`, and `docs` can be `false` or an `object`. If they are an empty object its the same as true but they can have [individual `src` folder settings](https://github.com/videojs/spellbook/blob/master/src/utils/get-config.js#L76-L91). This is outside the scope of the generator, but its possible.
*  I also noticed, looking back at the code, that there is [one Array setting too](https://github.com/videojs/spellbook/blob/master/src/utils/get-config.js#L66).

PR #102, opened 2016-11-04
No Babel

### LINKED ISSUE #93: Require Node 4.4.0 and higher

misteroneill:
Fixes #93 and removes Babel entirely from the generator. There's no longer any need to build in order for code to pass videojs-standard.


bin/videojs-generator-cleanup.js
```diff
@@ -1,20 +1,34 @@
...
-var chalk = require('chalk');
-var fs = require('fs-extra');
-var _ = require('lodash');
-var path = require('path');
-var semver = require('semver');
-var sh = require('shelljs');
+'use strict';
 
-var configs = {
+/* eslint-env node */
```
BrandonOCasey: isn't this line a part of videojs-standard already?

package.json
```diff
@@ -14,28 +14,20 @@
...
-    "build:docs": "doctoc README.md docs/",
-    "build:generator": "babel --ignore templates/**/* src/app -d generators/app && npm run build:generator:copy",
-    "build:generator:copy": "node -e \"require('shelljs').cp('-r', 'src/app/templates', 'generators/app')\"",
-    "build:test": "babel src/test -d es5-test",
     "lint": "vjsstandard",
-    "prepublish": "npm run build",
-    "pretest": "npm run lint && npm run build",
-    "test": "mocha es5-test/*.test.js",
+    "pretest": "npm run lint",
+    "test": "mocha test/*.test.js",
```
BrandonOCasey: we may want to add back a prepublish for test/lint

generators/app/index.js
```diff
@@ -1,14 +1,16 @@
...
-import path from 'path';
-import tsmlj from 'tsmlj';
-import yeoman from 'yeoman-generator';
-import yosay from 'yosay';
-import shelljs from 'shelljs';
-
-import {PREFIX} from './constants';
-import packageJSON from './package-json';
-import * as validators from './validators';
+'use strict';
```
BrandonOCasey: I think all node modules have this on by default. So I don't think we need to specify this here or in the other files.

misteroneill: Node 4.4 complains about using `const` and `let` outside of strict mode.

BrandonOCasey: ok

BrandonOCasey:
I think we want to add the node 4.4+ requirement to the package.json template file

PR #103, opened 2016-11-11
docs: Note that plugin is created in current directory

mister-ben:
I repeatedly fail to remember whether the plugin is created in the current directory, or a directory is created (the former).

Adds a note to the readme to save (my) confusion.


BrandonOCasey:
LGTM, I have had this issue as well. Maybe we should have an issue to allow the directory to be passed in.

PR #108, opened 2017-02-09
Video.js 5/6 cross-compatibility

misteroneill:
This updates the generator v2 to output a plugin that works with both Video.js 5 and 6. This has been tested and verified.

A subsequent PR will be opened to allow outputting an advanced plugin instead of a basic one.

### DIFF
--- a/src/app/templates/_.travis.yml
+++ b/src/app/templates/_.travis.yml
@@ -24,7 +24,7 @@ env:
   - VJS=5
   - VJS=6
 addons:
-  firefox: latest-esr
+  firefox: latest
   apt:
     sources:
     - google-chrome
--- a/src/app/templates/_.travis.yml
+++ b/src/app/templates/_.travis.yml
@@ -1,13 +1,32 @@
 sudo: false
+dist: trusty
 language: node_js
 node_js:
   - 'node'
-  - '4.4'
-  - '0.12'
-  - '0.10'
-
+  - 'lts/argon'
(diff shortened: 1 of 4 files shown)


src/app/templates/_.travis.yml
```diff
@@ -1,13 +1,32 @@
...
+      echo "video.js@$VJS ALREADY INSTALLED";
+    fi'
+  - export CHROME_BIN=/usr/bin/google-chrome
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+env:
+  - VJS=5
+  - VJS=6
+addons:
+  firefox: latest-esr
```
BrandonOCasey: I am not sure if this should be `latest` or `latest-esr`. I have been using `latest`, except in projects that already specify `latest-esr`.

src/app/templates/src/_plugin.js
```diff
@@ -3,6 +3,9 @@ import videojs from 'video.js';
 // Default options for the plugin.
 const defaults = {};
 
+// Cross-compatibility for Video.js 5 and 6.
+const registerPlugin = videojs.registerPlugin || videojs.plugin;
```
BrandonOCasey: Maybe we should include a commented out videojs.dom here too? Just for reference?

misteroneill: Sounds reasonable.

PR #121, opened 2017-04-06
use husky instead of ghooks

BrandonOCasey:
I don't think this is a major version bump, as it shouldn't actually break anything.

### DIFF
--- a/generators/app/package-json.js
+++ b/generators/app/package-json.js
@@ -8,8 +8,7 @@ const DEFAULTS = {
     'video.js': '^5.16.0'
   },
   devDependencies: {
-    'videojs-spellbook': '^3.0.0',
-    'ghooks': '^1.3.2'
+    'videojs-spellbook': '^3.0.0'
   }
 };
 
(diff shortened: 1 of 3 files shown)


generators/app/package-json.js
```diff
@@ -148,18 +147,29 @@ module.exports = (current, context) => {
   // In case ghooks was previously installed, but is now "none", we can
   // remove it from the package.json entirely.
   if (context.ghooks === 'none') {
-    delete result.devDependencies.ghooks;
+    delete result.devDependencies.ghoks;
```
misteroneill: Typo `ghoks` here.

generators/app/package-json.js
```diff
@@ -148,18 +147,29 @@ module.exports = (current, context) => {
...
+    if (result.config) {
+      delete result.config.ghooks;
+      if (Object.keys(result.config).length === 0) {
+        delete result.config;
       }
-    };
+    }
+
+    // delete old ghooks
+    delete result.devDependencies.ghoks;
```
misteroneill: `ghoks` again.

PR #131, opened 2017-08-04
fix: start should not build before starting, as watch will do that

BrandonOCasey:


gkatsev:
I think this was because originally, the watch didn't necessarily build everything.

PR #134, opened 2017-08-18
fix: set global as external during watch, to prevent an error

BrandonOCasey:
The first build of js-modules passes, but subsequent builds fail due to a [bug in rollup that has yet to be resolved](https://github.com/rollup/rollup/issues/982). Since we don't bundle global in the module build anyway we can set it as external and stop the error from happening.


misteroneill:
Weird, I haven't encountered this at all.

misteroneill:
Nevermind, I have, misunderstood at first. :)

PR #136, opened 2017-10-23
fix: tick 2 for player ready

BrandonOCasey:
https://github.com/videojs/video.js/pull/4665


gkatsev:
We probably merged this a bit early but it doesn't hurt

PR #152, opened 2018-02-03
feat: Remove Sass support option.

misteroneill:
BREAKING CHANGE: Generated projects no longer support Sass out of the box.


misteroneill:
Yeah, I think we should offer something - probably PostCSS or, maybe, LESS.

gkatsev:
postcss, or just nothing and let users choose what they want.
I do agree that it should make sure not to remove sass when upgrading and it is currently being used. Probably should ask "are you currently using sass? Do you want to continue using sass?"

gkatsev:
we can always add postcss later if we want to go that route.

misteroneill:
I like that idea. We can add PostCSS as a minor at some point. Part of the idea here is that the generator's output should be light-weight.

The generator is generally non-destructive, so I don't think it will remove Sass if it's in use. I'll verify, though.

PR #153, opened 2018-02-03
feat: Support Video.js 7 only and move to `peerDependencies`.

### LINKED ISSUE #146: Make video.js a peerDependency in package.json
The video.js plugin is currently listed as a standard dependency but this can cause issues when using the plugin with tools like Webpack as it can cause a second copy of video.js to be included if you are using multiple plugins. Typically plugins should be declared as peerDependencies so that each plugin doesn't get it's own separate copy/instance of video.js. See this post for more details: https://nodejs.org/en/blog/npm/peer-dependencies/

misteroneill:
Fixes #146

BREAKING CHANGE: This removes cross-version support between Video.js 5 and 6.

### DIFF
--- a/generators/app/package-json.js
+++ b/generators/app/package-json.js
@@ -159,7 +159,7 @@ const packageJSON = (current, context) => {
       'prepublish': 'not-in-install && npm run build || in-install',
       'start': 'npm-run-all -p server watch',
       'server': 'karma start scripts/karma.conf.js --singleRun=false --auto-watch --no-browsers',
-      'pretest': 'npm run lint',
+      'pretest': 'npm-run-all lint',
       'test': 'karma start scripts/karma.conf.js',
       'preversion': 'npm test',
       'version': 'node scripts/version.js',
@@ -223,6 +223,7 @@ const packageJSON = (current, context) => {
   }
 
   if (context.css) {
+    result.scripts.pretest += ' postclean build:css';
 
     _.assign(result.scripts, {
       'build:css': scriptify('postcss --verbose -o dist/%s.css --config scripts/postcss.config.js src/plugin.css'),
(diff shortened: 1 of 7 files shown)


gkatsev:
Personally, I think that peerDependencies are completely broken and having permissive ranges in the dependencies allow bundlers to work better without the pain of dealing with peerDependencies.

gkatsev:
I'm not inherently against it but I've had bad experience with peerDeps and would need to be convinced that the current implementations don't screw things up worse than just permissive semver ranges for dependencies and a good bundler along with deduping.

misteroneill:
Yeah, I'm not inherently for it or against it. I'd like to experiment with it at some point, but that's a slightly time-consuming task and we have other priorities. There's no rush on shipping a 6.0 yet, though.

generators/app/package-json.js
```diff
@@ -24,8 +24,10 @@ const getGeneratorVersions = (pkgList) => pkgList.reduce((acc, pkgName) => {
 }, {});
 
 const DEFAULTS = {
-  dependencies: getGeneratorVersions(['global', 'video.js']),
+  dependencies: getGeneratorVersions(['global']),
+  peerDependencies: getGeneratorVersions(['video.js']),
   devDependencies: getGeneratorVersions([
+    'video.js',
```
BrandonOCasey: It needs to be a peer and dev dependency so that we can use it for testing.

generators/app/templates/_index.html
```diff
@@ -20,7 +20,7 @@
   <script>
     (function(window, videojs) {
       var player = window.player = videojs('<%= pluginName %>-player');
-      player.<%= pluginFunctionName %>();
+      var <%= pluginFunctionName %> = window.<%= pluginFunctionName %> = player.<%= pluginFunctionName %>();
```
BrandonOCasey: Added this small fix in here, as I think its nice to be able to use you plugin on window right away.

generators/app/package-json.js
```diff
@@ -250,6 +249,12 @@ const packageJSON = (current, context) => {
   // here because the videojs-languages package will create the destination
   // directory if needed.
   if (context.lang) {
+    if (!context.docs) {
```
BrandonOCasey: Another small fix to add `build:lang`, and `build:css` to `pretest`. Since we could need those for tests. js building is already done by karma itself.

BrandonOCasey:
After testing with this, I don't think we have much to gain by making it a peer dependency other than complexity. I really think that we should just recommend that bundlers alias away video.js to a global.

PR #154, opened 2018-02-03
feat: Allow choosing between a basic and an advanced plugin template.

misteroneill:


generators/app/templates/src/_plugin-advanced.js
```diff
@@ -0,0 +1,47 @@
...
+   *         An optional options object.
+   *
+   *         While not a core part of the Video.js plugin architecture, a
+   *         second argument of options is a convenient way to accept inputs
+   *         from your plugin's caller.
+   */
+  constructor(player, options) {
+    super(player);
+
+    this.options = videojs.mergeOptions(defaults, options);
```
BrandonOCasey: Maybe add a `this.player = player` here too?

misteroneill: That happens implicitly in `super()`. Maybe I'll add a comment indicating that `this.player` is available.

generators/app/constants.js
```diff
@@ -48,6 +53,7 @@ module.exports = {
     husky: 'lint',
     lang: false,
     license: LICENSE_DEFAULT,
+    pluginType: 'basic',
```
gkatsev: I think the default should be "advanced".

generators/app/templates/src/_plugin-advanced.js
gkatsev: We should link to the docs and blog posts and such.

PR #155, opened 2018-02-04
refactor: Consolidate disparate Rollup configs into a single Rollup config.

misteroneill:


misteroneill:
Yeah, suppose we could do that. Sounds good.

BrandonOCasey:
see #166

BrandonOCasey:
* I added the banner to a js file, so that postcss can use it when the time comes
* I made karma do its own rollup, so that we don't have to worry about test building
* I removed `test/index.html` in favor of the karma debug page. This will be improved more in #168

generators/app/templates/scripts/_banner.js
```diff
@@ -0,0 +1,3 @@
+const pkg = require('../package.json');
+
+module.exports = `/*! @name ${pkg.name} @version ${pkg.version} @copyright ${pkg.author} @license ${pkg.license} */`
```
BrandonOCasey: Do we want this on multiple lines like before? seems like a waste of space to me, but I don't feel strongly that it has to be done on the same line.

generators/app/templates/scripts/_server.js
```diff
@@ -3,11 +3,19 @@
 const http = require('http');
 const portscanner = require('portscanner');
 const nodeStatic = require('node-static');
+const url = require('url');
 
 const files = new nodeStatic.Server(process.cwd(), {cache: false});
 
 const server = http.createServer((request, response) => {
   response.setHeader('Cache-Control', 'no-cache,must-revalidate');
+  const requestUrl = url.parse('http://' + request.headers.host + request.url);
```
BrandonOCasey: Auto redirect `/test` to karma server

generators/app/templates/test/_karma.conf.js
```diff
@@ -11,7 +19,7 @@ module.exports = function(config) {
 
   // If no browsers are specified, we enable `karma-detect-browsers`
   // this will detect all browsers that are available for testing
-  if (!config.browsers.length) {
+  if (config.browsers !== false && !config.browsers.length) {
```
BrandonOCasey: passing `--no-browsers` causes this to be false rather than an empty array, and we want to respect `--no-browsers` for the `test:server`

PR #165, opened 2018-06-01
feat: add generator deps as optional for greenkeeper

BrandonOCasey:


.npmrc
```diff
@@ -0,0 +1 @@
+optional = false
```
BrandonOCasey: Don't install optional dependencies by default

PR #168, opened 2018-06-04
Feat/karma server only

BrandonOCasey:
Use the karma server and a middleware rather than our own. This gets us:

* A static file server like we have now.
* we dont have to guess what port karma is on, it will be the same as the dev server
* static files can be accessed from karma (outside of the test dir).
* we can remove portscanner as karma will do it for us
* one less script file


generators/app/templates/scripts/_karma.conf.js
```diff
@@ -1,11 +1,36 @@
...
   'qunit': 'QUnit',
   'qunitjs': 'QUnit',
   'sinon': 'sinon',
   'video.js': 'videojs'
 };
 
+const files = new nodeStatic.Server(process.cwd(), {cache: false});
+const StaticMiddlewareFactory = function(config) {
+  console.log(`> Dev server started at http://${config.listenAddress}:${config.port}/`);
+  return function(request, response, next) {
```
BrandonOCasey: A middleware to serve static files

generators/app/templates/scripts/_karma.conf.js
```diff
@@ -50,21 +75,27 @@ module.exports = function(config) {
     },
     detectBrowsers,
     reporters: ['dots'],
-    port: 9876,
+    port: 9999,
```
BrandonOCasey: default dev server port

generators/app/templates/scripts/_karma.conf.js
```diff
@@ -50,21 +75,27 @@ module.exports = function(config) {
...
     colors: true,
     autoWatch: false,
     singleRun: true,
     concurrency: Infinity,
     preprocessors: {
       'test/**/*.test.js': ['rollup']
     },
+    urlRoot: '/test/',
```
BrandonOCasey: change the karma url root to /test

generators/app/package-json.js
```diff
@@ -27,7 +27,6 @@ const DEFAULTS = {
     'mkdirp': '^0.5.1',
     'node-static': '^0.7.10',
     'npm-run-all': '^4.1.2',
-    'portscanner': '^2.1.1',
```
BrandonOCasey: karma does this for us.

BrandonOCasey:
The benefit is that we dont have to guess the karma server port (9876 by default but it will chanhe if there is more than one server). Also the karma server does port scanning on its own. 9999 becomes 10000 if 9999 is taken etc

BrandonOCasey:
other than that the middleware is the same as the old server

gkatsev:
Does karma print out the url/port? Does it also have a link to the debug page? Would be cool to add if it doesn't.

BrandonOCasey:
Karma prints out the url and port, but I also print it out in the middleware on init. It does not link to the debug page, but because of the client options we set for qunit even the main page will have debugging abilities. I will change the `index.html` to point to the debug page though.

PR #172, opened 2018-06-05
feat: postcss support

BrandonOCasey:


package.json
```diff
@@ -51,10 +51,12 @@
     "videojs-standard": "^6.0.0"
   },
   "optionalDependencies": {
+    "autoprefixer": "^8.6.0",
```
gkatsev: another one to consider is ~cssnext~ postcss-preset-env (https://preset-env.cssdb.org/) (apparently, cssnext has been deprecated)

generators/app/templates/scripts/_postcss.config.js
```diff
@@ -0,0 +1,39 @@
+const banner = require('./banner');
+const postcss = require('postcss');
+const path = require('path');
+const fs = require('fs');
+
+const browserList = ['defaults', 'ie 11'];
```
gkatsev: what about adding firefox ESR?

PR #180, opened 2018-06-11
fix: run tests on 6 and 7 and support greenkeeper lock

BrandonOCasey:


generators/app/package-json.js
```diff
@@ -225,7 +225,7 @@ const packageJSON = (current, context) => {
   }
 
   if (context.css) {
-    result.scripts.pretest += ' postclean build:css';
+    result.scripts.pretest += ' clean build:css';
```
BrandonOCasey: small fix that I found converting a project, we may as well do clean and postclean.

PR #185, opened 2018-06-25
feat: add teamcity & browserstack support

BrandonOCasey:


generators/app/templates/scripts/_karma.conf.js
```diff
@@ -1,62 +1,107 @@
...
-const path = require('path');
-const serve = serveStatic(
-  path.join(__dirname, '..'),
-  {index: ['index.html', 'index.htm']}
-);
 
+/* allow static files to be served, note karma will takeover the /test directory */
 const StaticMiddlewareFactory = function(config) {
-  console.log(`**** Dev server started at http://${config.listenAddress}:${config.port}/ *****`);
+  console.log(`**** Karma static file server enabled in project root *****`);
```
BrandonOCasey: small change here, config.port would be wrong if the port is in-use and has to be incremented.

generators/app/templates/scripts/_karma.conf.js
```diff
@@ -1,62 +1,107 @@
...
+
+  const serve = serveStatic(
+    config.basePath,
+    {index: ['index.html', 'index.htm']}
+  );
```
BrandonOCasey: We should server the basePath, rather than trying to determine our own.

generators/app/templates/scripts/_karma.conf.js
```diff
@@ -1,62 +1,108 @@
...
-const path = require('path');
-const serve = serveStatic(
-  path.join(__dirname, '..'),
-  {index: ['index.html', 'index.htm']}
-);
+const pkg = require('../package.json');
 
+/* allow static files to be served, note karma will takeover the /test directory */
 const StaticMiddlewareFactory = function(config) {
-  console.log(`**** Dev server started at http://${config.listenAddress}:${config.port}/ *****`);
```
gkatsev: is this info being printed elsewhere?

BrandonOCasey: The karma server prints it when it starts up. Oftentimes the address we were printing here is the address we were passed. This means that when port `9999` is in use and get incremented by karma to `10000`. we still print `9999`.

gkatsev: cool.

generators/app/index.js
```diff
@@ -305,16 +305,14 @@ module.exports = class extends Generator {
 
     const file = constants.LICENSE_FILES[this.config.get('license')];
 
-    if (!file) {
-      return;
+    if (file) {
```
BrandonOCasey: Fixes a bug where un-licensed would cause the package.json not to be copied.

generators/app/templates/scripts/_rollup.config.js
```diff
@@ -9,9 +9,15 @@ const resolve = require('rollup-plugin-node-resolve');
 const {uglify} = require('rollup-plugin-uglify');
 const {minify} = require('uglify-es');
 const pkg = require('../package.json');
+
+/* General Globals */
+const moduleName = '<%= moduleName %>';
+const pluginName = '<%= pluginName %>';
```
BrandonOCasey: Mostly I added all the main variables at the top of the file, so that users won't have to dig into the actual builds unless they really need to.

generators/app/templates/scripts/_karma.conf.js
```diff
@@ -1,62 +1,108 @@
...
   };
 };
 
-module.exports = function(config) {
-  const detectBrowsers = {
-    enabled: false,
-    usePhantomJS: false
-  };
+/* browsers to run on teamcity */
+const teamcityLaunchers = {
```
gkatsev: this seems a confusing name given that these browsers are set up to work with browserstack.

gkatsev:
Also, would've been nice if the rollup changes, karma changes, and the fix for package.json were all in separate PRs.

generators/app/templates/scripts/_karma.conf.js
BrandonOCasey: yeah it is maybe we should split it into `teamcityLaunchers`, and `browserstackLaunchers`

BrandonOCasey:
I will split up the prs

generators/app/templates/scripts/_karma.conf.js
gkatsev: yeah, I would just call these browserstackLaunchers and just use it in TC usecase below or something.

PR #186, opened 2018-06-26
feat: add postcss features to make migrating sass easier

BrandonOCasey:
* Added a way to flatten `calc()` calls (even though they are supported in IE 11)
* Make sure that `var()` calls are flattened as they are not supported in IE 11
* Print the time taken and remove verbose.


generators/app/package-json.js
```diff
@@ -229,7 +229,7 @@ const packageJSON = (current, context) => {
 
   if (context.css) {
     _.assign(result.scripts, {
-      'build:css': scriptify('postcss --verbose -o dist/%s.css --config scripts/postcss.config.js src/plugin.css'),
+      'build:css': scriptify('postcss -o dist/%s.css --config scripts/postcss.config.js src/plugin.css'),
```
BrandonOCasey: the verbose output for postcss cli is basically useless it just prints the file that was processed in color and nothing else.

generators/app/templates/scripts/_postcss.config.js
```diff
@@ -15,13 +18,26 @@ const printOutput = postcss.plugin('postcss-print-output', function(opts) {
...
-    console.log(`${relativeFrom} -> ${relativeTo}`);
+    console.log(`${relativeFrom} -> ${relativeTo} in ${getNow() - startTime}ms`);
+  };
+});
+
+/**
+ * A function to set the startTime of postcss so that
+ * we can print the time taken in the output.
+ */
+const setTime = postcss.plugin('postcss-set-time', function(opts) {
```
gkatsev: looks like there's a plugin that will do some of this already? https://github.com/postcss/postcss-devtools

BrandonOCasey: will test

BrandonOCasey:
re: calc on IE11, Only for minification, but the minifier will already do it so I will remove it.

generators/app/templates/scripts/_postcss.config.js
BrandonOCasey: This plugin makes the build much slower and also prints out debug information that would probably only be useful when actually developing postcss.

gkatsev: cool. Maybe at some point we can make that a more fully featured postcss plugin like time-grunt?

PR #194, opened 2018-07-13
fix: do not exclude node modules from babel

BrandonOCasey:
In order to get many of the size benefits from rollup/es6 code we want to use es6 versions of sub modules. Right now that is what we do, but it causes syntax errors (especially on ie 11) since anything in node_modules won't be converted to es5 and will remain es6.

### DIFF
--- a/generators/app/templates/scripts/_rollup.config.js
+++ b/generators/app/templates/scripts/_rollup.config.js
@@ -20,7 +20,9 @@ const banner = `/*! @name ${pkg.name} @version ${pkg.version} @license ${pkg.lic
 const primedPlugins = {
   babel: babel({
     babelrc: false,
-    exclude: 'node_modules/**',
+    // you may want to manually exclude non-es6 modules
+    // if your build is slow
+    exclude: [],
     presets: [
       ['env', {loose: true, modules: false, targets: {browsers: pkg.browserslist}}]
     ],


gkatsev:
do we have dependencies that ship es6 code hidden in es modules?

BrandonOCasey:
Any plugin that uses the generator that depends on another plugin that uses the generator will have this problem, because rollup prioritizes the `module` entry in `package.json` and uses that. Previously this wasn't a problem because our [`module`/`es` entry was actually being babeled down to `es5`](https://github.com/videojs/generator-videojs-plugin/blob/5.x/generators/app/templates/scripts/_modules.rollup.config.js#L35)

gkatsev:
Is the `.es.js` output using es6 code? The only non es5 code it should have is `import` statements (es modules stuff)

BrandonOCasey:
shouldn't the es module basically be the source code all bundled together without any transpiration to es5? For instance a class stays a class.

gkatsev:
no, the `pkg.module` should basically just be es module stuff, everything else should be compiled down to work on our standard browser support, so, IE11+ etc. We shouldn't be forcing users of our module to run babel on our stuff.

BrandonOCasey:
After further discussion it appears that the es dist should be babeled and I was wrong. https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features

Closing this in favor of #196 which brings that functionality back.

PR #200, opened 2018-07-17
fix: shim globals to fix requirejs

BrandonOCasey:
Still not sure if we should go with this or #197. Personally I lean towards the other one as it does not mess with any code except for requirejs.

### DIFF
--- a/generators/app/templates/_README.md
+++ b/generators/app/templates/_README.md
@@ -54,10 +54,25 @@ player.<%= pluginFunctionName %>();
 When using with RequireJS (or another AMD library), get the script in whatever way you prefer and `require` the plugin as you normally would:
 
 ```js
-require(['video.js', '<%= packageName %>'], function(videojs) {
-  var player = videojs('my-video');
+// configure the paths to the packages with requirejs
+requirejs.config({
+  paths: {
+    "videojs": "videojs-url",
+    "<%= pluginFunctionName %>": "plugin-url",
+  },
+});
 
-  player.<%= pluginFunctionName %>();
+// first require videojs (unless you included it as a script on the page)
+requirejs(['videojs'], function(videojs) {
+  // add videojs to window
+  window.videojs = videojs;
+
+  // then require this plugin, as it expects a global videojs on window.
+  requirejs(['<%= pluginFunctionName %>'], function(<%= pluginFunctionName %>) {
+    var player = videojs('my-video');
+
+    console.log(player.<%= pluginFunctionName %>);
+  });
 });
 ```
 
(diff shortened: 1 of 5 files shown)


generators/app/templates/scripts/_rollup.config.js
```diff
@@ -78,10 +59,52 @@ const externals = {
   ])
 };
 
+/* configuration for plugins */
+const primedPlugins = {
+  babel: babel({
+    babelrc: false,
```
BrandonOCasey: had to move this down so that shim has access to globals/externals

generators/app/templates/scripts/_rollup.config.js
```diff
@@ -78,10 +59,52 @@ const externals = {
...
+  resolve: resolve({browser: true, main: true, jsnext: true}),
+  uglify: uglify({output: {comments: 'some'}}, minify),
+  /* we have to shim externals to a string to work-around amd bugs in rollup */
+  shim: shim(externals.umd.reduce(function(newObj, name) {
+    // set the shim to the global or empty object
+    newObj[name] = globals.umd[name] || '{}';
+
+    // if the shim is not window, document, or empty object add window to the front of it.
+    if (newObj[name] !== 'window' && newObj[name] !== 'document' && newObj[name] !== '{}') {
+      newObj[name] = 'window.' + newObj[name];
```
gkatsev: we should probably do a check along the lines of
```
typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : {}
```
to get the global object to be defensive.

BrandonOCasey: done

generators/app/templates/_README.md
```diff
@@ -54,10 +54,25 @@ player.<%= pluginFunctionName %>();
 When using with RequireJS (or another AMD library), get the script in whatever way you prefer and `require` the plugin as you normally would:
 
 ```js
-require(['video.js', '<%= packageName %>'], function(videojs) {
-  var player = videojs('my-video');
+// configure the paths to the packages with requirejs
+requirejs.config({
+  paths: {
+    "videojs": "videojs-url",
+    "<%= pluginFunctionName %>": "plugin-url",
```
misteroneill: It seems wrong that we'd have to configure paths at all. I feel like a UMD build should "just work".

ony19161:
I am using requireJs with Backbone, for me it worked after amdfying videojs as following 
1. Create a new file named videoJsAmd, and inside the file added the following code
define(['libs/videojs/video.min'], function(videojs){
    return videojs;
});
2. Referenced the videoJsAmd in require.confg as following 
require.config({
    baseUrl: "/ClientCode/Scripts/",
    shim: {
        "libs/backbone/backbone": {
            deps: ["libs/underscore/underscore", "libs/jquery/jquery.min"],
            exports: "Backbone"
        },
        "libs/jquery.tmpl/jquery.tmpl.min": ["libs/jquery/jquery.min"]
    },
    paths: {
        jqueryAMD: 'libs/jquery/jquery.amd',
        underscoreAMD: 'libs/underscore/underscore.amd',
        backboneAMD: 'libs/backbone/backbone.amd',
        videoJsAMD: 'libs/videoJsAmd'
    }
});

PR #202, opened 2018-07-18
Greenkeeper/rollup 0.63.2

BrandonOCasey:
Closes #190


BrandonOCasey:
This does not change anything from a diff and appears to work the same as the previous version.

PR #204, opened 2018-07-19
Chore/update videojs standard

BrandonOCasey:


test/libs.js
```diff
@@ -12,14 +12,12 @@ const FILES = {
     'scripts/rollup.config.js',
     'scripts/version.js',
     'src/plugin.js',
-    'test/index.html',
```
BrandonOCasey: we don't make this file anymore, since 6.0.0

test/libs.js
```diff
@@ -12,14 +12,12 @@ const FILES = {
...
     'scripts/karma.conf.js',
     'test/plugin.test.js',
     '.editorconfig',
     '.gitignore',
     '.npmignore',
     '.nvmrc',
-    'scripts/jsdoc.json',
```
BrandonOCasey: We only make this file for docs, seems like the tests were majorly broken until now.

test/libs.js
```diff
@@ -53,8 +51,9 @@ const GENERATOR_PATH = path.join(__dirname, '../generators/app');
  * @return   {Array}
  *           An array of filenames.
  */
-const fileList = () =>
-  _.union.apply(_, _.toArray(arguments).map(str => FILES[str]));
+const fileList = function() {
+  return _.union.apply(_, _.toArray(arguments).map(str => FILES[str]));
```
BrandonOCasey: arguments was undefined for an arrow function, which turned this to an empty array, so it was searching for nothing...

README.md
```diff
@@ -1,6 +1,10 @@
 # generator-videojs-plugin
 
+[![Build Status](https://travis-ci.org/videojs/generator-videojs-plugin.svg?branch=master)](https://travis-ci.org/videojs/generator-videojs-plugin)
 [![Greenkeeper badge](https://badges.greenkeeper.io/videojs/generator-videojs-plugin.svg)](https://greenkeeper.io/)
```
BrandonOCasey: Our now standard badges!

PR #209, opened 2018-08-16
feat: replace version script with not-prerelase

BrandonOCasey:


generators/app/package-json.js
```diff
@@ -164,7 +164,8 @@ const packageJSON = (current, context) => {
       'pretest': 'npm-run-all lint build',
       'test': 'karma start scripts/karma.conf.js',
       'preversion': 'npm test',
-      'version': 'node scripts/version.js',
+      'version': 'is-prerelease || npm run update-changelog && git add CHANGELOG.md',
```
BrandonOCasey: is-prerelease exits with a true value if the current version is a `prerelease`. If it exits false we run the code after the `||`

PR #220, opened 2018-08-28
feat: add code coverage reporting by updating deps

BrandonOCasey:


scripts/run-integration-tests.js
```diff
@@ -12,7 +12,7 @@ helpers.run(libs.GENERATOR_PATH)
   })
   .withOptions(libs.options())
   .withPrompts({
-    name: 'REPLACE_ME',
+    name: 'integration-test',
```
BrandonOCasey: When this test was failing it threw me off, because the name is REPLACE_ME

generators/app/package-json.js
```diff
@@ -143,6 +144,7 @@ const packageJSON = (current, context) => {
       'server': 'karma start scripts/karma.conf.js --singleRun=false --auto-watch --no-browsers',
       'pretest': 'npm-run-all lint build',
       'test': 'karma start scripts/karma.conf.js',
+      'posttest': 'shx cat test/dist/coverage/text.txt',
```
BrandonOCasey: If we actual report this in karma we get coverage for every browser that runs, which is almost always the same.

generators/app/package-json.js
```diff
@@ -37,7 +37,8 @@ const DEFAULTS = {
     'videojs-generate-karma-config',
     'not-prerelease',
     'sinon',
-    'videojs-standard'
+    'videojs-standard',
+    'npm-merge-driver'
```
BrandonOCasey: super small module that makes rebasing projects that use `package-lock` super easy. Basically it will merge package-locks when they conflict much faster than regenerating them ourselves.

misteroneill:
Approved provided tests pass.

PR #230, opened 2018-09-07
feat: use es-check to verify dist

BrandonOCasey:
Checks to make sure that our dist files are not using unexpected syntax

### DIFF
--- a/generators/app/package-json.js
+++ b/generators/app/package-json.js
@@ -41,7 +41,8 @@ const DEFAULTS = {
     'npm-merge-driver-install',
     'husky',
     'lint-staged',
-    'pkg-ok'
+    'pkg-ok',
+    'es-check'
   ])
 };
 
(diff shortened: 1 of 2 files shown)


generators/app/package-json.js
```diff
@@ -142,7 +143,12 @@ const packageJSON = (current, context) => {
       'clean': 'shx rm -rf ./dist ./test/dist',
       'postclean': 'shx mkdir -p ./dist ./test/dist',
       'lint': 'vjsstandard',
-      'prepublish': 'not-in-install && npm run build && pkg-ok || in-install',
+      'prepublish': 'not-in-install && npm run verify || in-install',
+      'preverify': 'npm run build',
```
gkatsev: running the build as a preverify seems weird. I think it would be better to keep it in prepublish as a serial run via npm-run-all `run-s build verify`.

generators/app/package-json.js
```diff
@@ -142,7 +143,12 @@ const packageJSON = (current, context) => {
...
+      'verify': 'npm-run-all -p verify:*',
+      'verify:pkg': 'pkg-ok',
+      'verify:es5': 'es-check es5 \'dist/!(*.es.js|*.css)\'',
+      'verify:es6': 'es-check --module true es6 dist/*.es.js',
```
gkatsev: the `es.js` file should be es5 aside from the usage of esmodules.

Also, I find it kind of weird that the flag takes a boolean rather than it being a boolean flag like `--no-module` and `--module`. 🤷‍♀️

generators/app/package-json.js
BrandonOCasey: Well in order to run verify manually (if anyone ever does that). we will have to have build happen first. That is why I did a `preverify`

generators/app/package-json.js
BrandonOCasey: Yeah its weird I am working on a few issues I found with the author over on the repo.

generators/app/package-json.js
gkatsev: Right, though, I would still find it confusing and weird that prepublish doesn't call build directly.
Maybe it should be `run-s build verify:*`? That way, it doesn't run the build again.

PR #232, opened 2018-09-12
feat: Use videojs-generator-verify to verify new packages

BrandonOCasey:
Depends on https://github.com/videojs/videojs-generator-verify/pull/1

Closes #230

### DIFF
--- a/generators/app/package-json.js
+++ b/generators/app/package-json.js
@@ -28,7 +28,6 @@ const DEFAULTS = {
   devDependencies: getGeneratorVersions([
     'conventional-changelog-cli',
     'conventional-changelog-videojs',
-    'in-publish',
     'karma',
     'npm-run-all',
     'rollup',
@@ -41,7 +40,7 @@ const DEFAULTS = {
     'npm-merge-driver-install',
     'husky',
     'lint-staged',
-    'pkg-ok'
+    'videojs-generator-verify'
   ])
 };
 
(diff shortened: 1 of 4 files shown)


generators/app/package-json.js
```diff
@@ -142,7 +141,7 @@ const packageJSON = (current, context) => {
       'clean': 'shx rm -rf ./dist ./test/dist',
       'postclean': 'shx mkdir -p ./dist ./test/dist',
       'lint': 'vjsstandard',
-      'prepublish': 'not-in-install && npm run build && pkg-ok || in-install',
+      'prepublishOnly': 'npm run build && vjsverify',
```
BrandonOCasey: not-in-install was not working correctly for me with `npm@6`, this may be something we have to update our projects with right away.

scripts/run-integration-tests.js
```diff
@@ -3,8 +3,12 @@
 const helpers = require('yeoman-test');
```
BrandonOCasey: I re-worked this file a lot, we now make sure that the `version`, `ci`, `install` and `prepublishOnly` scripts will work by using the various commands. In the future we should probably add `audit` here as well.

generators/app/package-json.js
misteroneill: Can `not-in-install` be fixed for npm@6 rather than opening a bunch more PRs?

BrandonOCasey: good question let me check

PR #233, opened 2018-09-12
feat: generate package-lock for faster install

BrandonOCasey:
Generate a package lock for plugin to save ~70s on initial install, even when deps have changed by selecting other options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "test": "npm-run-all test:*",
     "test:unit": "mocha test/*.test.js",
     "test:integration": "node scripts/run-integration-tests.js",
-    "preversion": "npm run test",
+    "preversion": "npm-run-all generate-pkg-lock test",
+    "generate-pkg-lock": "node scripts/generate-pkg-lock.t js && git add generators/app/templates/_package-lock.json",
     "version": "is-prerelease || npm run update-changelog && git add CHANGELOG.md",
     "update-changelog": "conventional-changelog -p videojs -i CHANGELOG.md -s",
     "prepublish": "not-in-publish || vjsverify --skip-es-check"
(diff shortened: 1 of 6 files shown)


gkatsev:
Interesting idea. With this, we can even run `npm ci` the very first time (maybe with a `--install` flag) which will make that initial install even faster.

BrandonOCasey:
Yeah we could, we would just have to version check npm, let me see if yeoman can do it

misteroneill:
I think you could simply skip the `npmInstall` call and manually invoke `npm ci` - if I'm understanding everything correctly.

Really cool idea! 👍

package.json
```diff
@@ -26,7 +26,8 @@
     "test": "npm-run-all test:*",
     "test:unit": "mocha test/*.test.js",
     "test:integration": "node scripts/run-integration-tests.js",
-    "preversion": "npm run test",
+    "preversion": "npm-run-all generate-pkg-lock test",
+    "generate-pkg-lock": "node scripts/generate-pkg-lock.t js && git add generators/app/templates/_package-lock.json",
```
BrandonOCasey: generate package lock and add it every version, I think this makes sense to me.

scripts/run-integration-tests.js
```diff
@@ -12,8 +12,9 @@ let tempDir;
 helpers.run(libs.GENERATOR_PATH)
   .inTmpDir(function(dir) {
     tempDir = dir;
+    console.log(`Generating Project in ${tempDir}`);
   })
-  .withOptions(libs.options())
+  .withOptions(Object.assign(libs.options(), {skipInstall: false}))
```
BrandonOCasey: I removed the install steps below, instead we do the builtin yeoman install

PR #234, opened 2018-09-13
fix: The --hurry option

### LINKED ISSUE #178: --hurry is broken
Seems like using variables in templates break when we do not have the prompt. Somewhere we are not setting all of the defaults correctly.

BrandonOCasey:
Closes #178


generators/app/constants.js
```diff
@@ -44,11 +44,11 @@ module.exports = {
   LICENSE_DEFAULT,
 
   PROMPT_DEFAULTS: {
-    docs: false,
-    css: false,
+    docs: true,
+    css: true,
```
BrandonOCasey: We can drop these changes here if we want, but I think that we want to offer the full toolset for --hurry.

generators/app/index.js
```diff
@@ -242,6 +246,7 @@ module.exports = class extends Generator {
    */
   prompting() {
     if (this.options.skipPrompt) {
+      _.assign(this._preconfigs, this._getPromptDefaults());
```
BrandonOCasey: Previously we did not set defaults for `--hurry` so everything was just undefined.

PR #244, opened 2018-10-04
feat: use vjsverify with tests

BrandonOCasey:


generators/app/package-json.js
```diff
@@ -137,11 +137,13 @@ const packageJSON = (current, context) => {
       'clean': 'shx rm -rf ./dist ./test/dist',
       'postclean': 'shx mkdir -p ./dist ./test/dist',
       'lint': 'vjsstandard',
-      'prepublishOnly': 'npm run build && vjsverify',
+      'prepublishOnly': 'vjsverify --verbose',
```
BrandonOCasey: Do we still want to build here? vjsverify should determine if things are built or not, and in almost all cases they will be since you do `npm version` before `npm publish`

gkatsev: Probably, I can definitely see cases where you'd want to be able to publish but you aren't going to run `npm version` directly, at least.

PR #246, opened 2018-10-09
fix: require node 8 in package.json

BrandonOCasey:


gkatsev:
fwiw, engines is basically never enforced.

BrandonOCasey:
What about adding: https://www.npmjs.com/package/please-upgrade-node

gkatsev:
That probably works.

misteroneill:
Was the index.js -> generator.js change just a rename? What was the actual change, if any?

BrandonOCasey:
The change was a rename just to make sure that the code will actually run in node < 8 since syntax like `class` won't be supported. I tested it on node 4 and 6 and it appears to exit as expected.

misteroneill:
Sounds good, thanks.

Issue #266, opened 2018-12-07
npm install not working because of event-stream@3.3.6


nicu-chiciuc:
I created a new folder, ran `yo videojs-plugin --skip-install` then `npm install`. I'm getting this error:
```
npm ERR! code E404
npm ERR! 404 Not Found: event-stream@3.3.6
```
This is because that version of the package was removed from npm because of the incident that happened in November: (https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident)

I think this should definitely fixed ASAP

nicu-chiciuc:
So the bad dependencies were these:
`npm-run-all` -> `ps-tree`-> `event-stream`
This was resolved here: https://github.com/mysticatea/npm-run-all/issues/149
I've updated `npm-run-all` from `4.1.3` to `4.1.5` and the install went through (although there are 2 high vulnerabilities, but this is another issue).

misteroneill:
Fixed by #267

PR #267, opened 2018-12-10
Updated npm-run-all from 4.1.3 to 4.1.5 so that event-stream@3.3.6 not used.

nicu-chiciuc:
The full problem is explained in [issues/266](https://github.com/videojs/generator-videojs-plugin/issues/266).
In short, `npm install` wasn't working since the dependecy `npm-run-all` was indirectly depending on `event-stream@3.3.6` which was removed from npm because of security reasons.

### DIFF
--- a/plugin/package.json
+++ b/plugin/package.json
@@ -22,7 +22,7 @@
     "lint-staged": "^8.0.0",
     "not-prerelease": "^1.0.1",
     "npm-merge-driver-install": "^1.0.0",
-    "npm-run-all": "^4.1.3",
+    "npm-run-all": "^4.1.5",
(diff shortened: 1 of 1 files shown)


misteroneill:
Uh, weird. This is already done in #263 - I wonder why it didn't register as a conflict? Strange.

misteroneill:
Oh, looks like #263 correctly updated the root package files, but this is updating the `plugin` directory. I'm not sure why that directory exists, though...

misteroneill:
I see. It looks like `plugin/package.json` exists to populate the dependencies for plugins. #263 fixed the root `package.json`, but that doesn't help generated plugins.

PR #289, opened 2019-07-15
feat: Update to videojs-generate-rollup-config ~5.0.0, use env options, and `@babel/runtime`

brandonocasey:
Use new environment variable options from `videojs-generate-rollup-config`@`~4.0.0` and `@babel/runtime` from `videojs-generate-rollup-config`@`~5.0.0`.

See: https://github.com/videojs/videojs-generate-rollup-config/blob/master/CHANGELOG.md

Depends upon: https://github.com/videojs/generator-helpers/pull/2


generators/app/package-json.js
```diff
@@ -124,18 +124,16 @@ const packageJSON = (current, context) => {
       'ie 11'
     ],
     'scripts': _.assign({}, current.scripts, {
-      'prebuild': 'npm run clean',
```
brandonocasey: Previously we used to do this prebuild, but since we have multiple different types of "build" now, it doesn't make sense.

generators/app/package-json.js
```diff
@@ -124,18 +124,16 @@ const packageJSON = (current, context) => {
...
-      'build': 'npm-run-all -p build:*',
+      'build-test': "npm-run-all -s clean -p 'build:!(js)' 'build:js -- --environment TEST_BUNDLE_ONLY'",
+      'build-prod': "npm-run-all -s clean -p 'build:!(js)' 'build:js -- --environment NO_TEST_BUNDLE'",
+      'build': 'npm-run-all -s clean -p build:*',
       'build:js': 'rollup -c scripts/rollup.config.js',
-      'clean': 'shx rm -rf ./dist ./test/dist',
-      'postclean': 'shx mkdir -p ./dist ./test/dist',
+      'clean': 'shx rm -rf ./dist ./test/dist && shx mkdir -p ./dist ./test/dist',
```
brandonocasey: May as well do it in one command

generators/app/package-json.js
```diff
@@ -124,18 +124,16 @@ const packageJSON = (current, context) => {
       'ie 11'
     ],
     'scripts': _.assign({}, current.scripts, {
-      'prebuild': 'npm run clean',
-      'build': 'npm-run-all -p build:*',
+      'build-test': "npm-run-all -s clean -p 'build:!(js)' 'build:js -- --environment TEST_BUNDLE_ONLY'",
+      'build-prod': "npm-run-all -s clean -p 'build:!(js)' 'build:js -- --environment NO_TEST_BUNDLE'",
```
brandonocasey: `build:!(js)` is run all build commands except for js, then we run it separately with an environment variable.

generators/app/package-json.js
```diff
@@ -124,18 +124,16 @@ const packageJSON = (current, context) => {
...
+      'build': 'npm-run-all -s clean -p build:*',
       'build:js': 'rollup -c scripts/rollup.config.js',
-      'clean': 'shx rm -rf ./dist ./test/dist',
-      'postclean': 'shx mkdir -p ./dist ./test/dist',
+      'clean': 'shx rm -rf ./dist ./test/dist && shx mkdir -p ./dist ./test/dist',
       'lint': 'vjsstandard',
-      'prepublishOnly': 'npm-run-all build test:verify',
+      'prepublishOnly': 'npm-run-all build-prod && vjsverify --verbose',
```
brandonocasey: `verify` can no longer be run on `test` so it only makes sense in pre-publish.

generators/app/package-json.js
gkatsev: I didn't know npm-run-all supported segation like that, nice.

plugin/package.json
```diff
@@ -10,18 +10,19 @@
   "license": "ISC",
   "dependencies": {
     "global": "^4.3.2",
-    "video.js": "^6 || ^7"
+    "video.js": "^6 || ^7",
+    "@babel/runtime": "^7.5.5"
   },
   "devDependencies": {
-    "@videojs/generator-helpers": "~1.1.1",
+    "@videojs/generator-helpers": "~1.2.0",
```
brandonocasey: This provides `cross-env`

generators/app/package-json.js
misteroneill: Although, I don't think`&&` will work on Windows, will it?

PR #302, opened 2020-03-24
chore: update videojs-generate-rollup-config to 5.0.2

gkatsev:


plugin/package.json
```diff
@@ -22,7 +22,7 @@
     "sinon": "^7.2.2",
     "videojs-generate-karma-config": "~5.3.1",
     "videojs-generate-postcss-config": "~2.1.0",
-    "videojs-generate-rollup-config": "~5.0.1",
+    "videojs-generate-rollup-config": "^5.0.2",
```
brandonocasey: should probably stay `~`

PR #307, opened 2021-04-30
feat(BREAKING): switch to node 14 and only test on safari 12/14, chrome, and firefox

brandonocasey:
node 8/10 are dead, 12 is very old, and 16 is very new. Let's make the jump to 14.


generators/app/templates/.github/workflows/_ci.yml
```diff
@@ -0,0 +1,77 @@
+name: ci
```
gkatsev, line 1: do we want this here? Since it's sort of specific to Videojs projects only and not everyone may want this. Plus, it'll be yet another place we have to maintain this file.

plugin/package.json
```diff
@@ -9,24 +9,24 @@
   "repository": "https://github.com/videojs/generator-videojs-plugin",
   "license": "ISC",
   "dependencies": {
-    "@babel/runtime": "^7.11.2",
+    "@babel/runtime": "^7.14.0",
     "global": "^4.4.0",
-    "video.js": "^6 || ^7"
+    "video.js": "^7.12.1"
```
gkatsev, line 14: I think this should stay the same

generators/app/templates/.github/workflows/_ci.yml
brandonocasey: Maybe, but we already included the travis config previously and didn't add it to the project if it was closed source. Should we fetch it on install?

plugin/package.json
brandonocasey: I'm not sure that video.js 6 with support node 14

generators/app/templates/.github/workflows/_ci.yml
gkatsev: downloading it from the .github repo sounds like a good idea. That way we don't have to maintain the script in yet another location.

plugin/package.json
gkatsev: I don't see that as an issue. The node support in Video.js is mostly in running the build itself. You definitely should be able to require vjs 6 in node 14 (and have it be a no-op, as it is now).
Ideally, this would stay the same until someone updates their plugin to not support IE11 and depend on Video.js 8.

generators/app/templates/.github/workflows/_ci.yml
brandonocasey: ok I have it downloading now.

plugin/package.json
brandonocasey: Reverted.

gkatsev (approved):

PR #309, opened 2021-05-26
fix: remove obsolete v5 compatiblity

mister-ben:
The dependency in package.json is `"^6 || ^7"` so the v5 compatiblity is unnecessary.


brandonocasey (approved):
👍🏻

